PR #56, opened 2022-03-24
Fix broken update calls when DomainReload disabled

mbaker3:
When [DomainReload](https://docs.unity3d.com/Manual/DomainReloading.html) is disabled static values aren't cleared between editor play sessions. The `AbstractUnityUpdateSourceMonoBehaviour` instances were being destroyed by the editor but their `AbstractUnityUpdateSource` owners weren't. Since the `AbstractUnityUpdateSource` weren't being disposed they weren't being unregistered from `UpdateHandleSourcesManager` and were delivered during the next play session without their MonoBehaviour counterpart to hook into and pump update events.

**Fix:**
 - Add `AbstractUnityUpdateSourceMonoBehaviour.OnDisposing` event
 - `AbstractUnityUpdateSource` detects `AbstractUnityUpdateSourceMonoBehaviour` and disposes self in editor if currently exiting play mode and wasn't already being disposed.
    - If not in play mode an error is logged because it's a problem if `AbstractUnityUpdateSourceMonoBehaviour`'s are getting destroyed without going through `AbstractUnityUpdateSource.Dispose()`

Unity update sources will fail to dispatch update events when the editor is played without a domain reload occurring.
**When domain reload is enabled** it is reloaded between every play session and there is no issue.
**When domain reload is disabled** it is only reloaded if the editor has recompiled.

The update sources dispatch an event that they're being disposed. If the editor is currently exiting play mode the `AbstractUnityUpdateSource` will dispose itself as well which, in turn, unregisters the instance from `UpdateHandleSourcesManager`.

`AbstractUnityUpdateSource` will now also log an error when its `MonoBehaviour` counterpart is disposed when the editor isn't exiting play mode and it's not initiated by the `AbstractUnityUpdateSource`. This acts as an assert against an unexpected state and will help developers track down incorrect behaviour if the `AbstractUnityUpdateSourceMonoBehaviour` gets destroyed accidentally.

The update system does not work consistently when Domain Reload is disabled.

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/DelayedExecution/UpdateSources/AbstractUnityUpdateSource.cs
+++ b/Scripts/Runtime/DelayedExecution/UpdateSources/AbstractUnityUpdateSource.cs
@@ -51,7 +51,7 @@ namespace Anvil.Unity.DelayedExecution
 #endif
 
             // This usually happens when m_UpdateSourceGameObject or one of its parents
-            // - Is accidentally destroyed by another process desroying GameObjects in the hierary.
+            // - Is accidentally destroyed by another process destroying GameObjects in the hierarchy.
             // - GameObject.DontDestroyOnLoad was not applied to the game object for some reason.
             Logger.Error($"{typeof(T)} was destroyed before its owner {this.GetType()}. Dispose must be called on the owner.");
         }
--- a/Scripts/Runtime/DelayedExecution/UpdateSources/AbstractUnityUpdateSource.cs
+++ b/Scripts/Runtime/DelayedExecution/UpdateSources/AbstractUnityUpdateSource.cs
@@ -8,31 +8,55 @@ namespace Anvil.Unity.DelayedExecution
     /// </summary>
     /// <typeparam name="T">Unity requires a MonoBehaviour for it's Update functions, this generic is of type
     /// <see cref="AbstractUnityUpdateSourceMonoBehaviour"/></typeparam>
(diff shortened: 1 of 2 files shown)


Scripts/Runtime/DelayedExecution/UpdateSources/AbstractUnityUpdateSource.cs
```diff
@@ -8,39 +8,63 @@ namespace Anvil.Unity.DelayedExecution
...
+            {
+                // Calling dispose explicitly allows this instance to be removed from 
+                // UpdateHandleSourcesManager. This is important when DomainReload is disabled.
+                Dispose();
+                return;
+            }
+#endif
+
+            // This usually happens when m_UpdateSourceGameObject or one of its parents
+            // - Is accidentally destroyed by another process desroying GameObjects in the hierary.
```
jkeon, line 54: typo - hierarchy

jkeon (approved):
Looks good, one minor typo

Scripts/Runtime/DelayedExecution/UpdateSources/AbstractUnityUpdateSource.cs
mbaker3: lol just suggest the change
```suggestion
            // - Is accidentally destroyed by another process destroying GameObjects in the hierarchy.
```

jkeon (approved):

PR #58, opened 2022-04-27
Feature - Add UnityLogListener

mbaker3:
Implementation of an `ILogListener` for the Unity logger.
Captures all logs emitted through `Debug.Log,Warn, Error,Assert, etc...` and redirects them through a Logger instance with a best effort made to select the original caller.

With this enabled setting Unity's console to single line is very viable. Now that caller context is included in the log message itself.

`Debug.Log, Warn, Error, Assert, etc...` calls do not pass through the `Log` system.

By default, `UnityLogListener` will be instantiated and direct all built in logging messages at runtime through `Log`. No developer interaction required. Just update to include these changes.

See issues resolved in https://github.com/decline-cookies/anvil-csharp-core/pull/80
Specifically, ECS has some false flag warnings during a valid operation that I wanted to suppress. While manipulating `Debug.unityLogger.logEnabled` would have worked this (combined with the link above) is a more comprehensive solution.

 - https://github.com/decline-cookies/anvil-csharp-core/pull/80

 - [ ] Yes
 - [x] No - Unless you really don't want debug messages sent through Log.

### DIFF
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
@@ -5,6 +5,10 @@ using StackFrame = System.Diagnostics.StackFrame;
 
 namespace Anvil.Unity.Logging
 {
+    /// <summary>
+    /// An implementation of <see cref="ILogListener"/> for the Unity <see cref="Debug"/> logger.
+    /// Captures all logging made through <see cref="Debug"/> and redirects them through a <see cref="Log.Logger"/>.
+    /// </summary>
     [DefaultLogListener]
     public sealed class UnityLogListener : ILogListener, UnityEngine.ILogHandler
     {
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
@@ -1,0 +1,99 @@
+using System;
+using Anvil.CSharp.Logging;
+using UnityEngine;
+using StackFrame = System.Diagnostics.StackFrame;
+
+namespace Anvil.Unity.Logging
+{
+    [DefaultLogListener]
+    public class UnityLogListener : ILogListener, UnityEngine.ILogHandler
+    {
+        private static UnityLogListener s_Instance;
+
+        // With the Burst compiler, it's common to disable the Domain Reload step when entering play mode to increase iteration time.
+        // https://docs.unity3d.com/Packages/com.unity.entities@0.17/manual/install_setup.html
+        // Because of this, statics will preserve state across play sessions.
+        // `RuntimeInitializeOnLoadMethod` ensures we reset the state on every play session.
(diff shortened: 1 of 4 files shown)


mbaker3:
Once the premise and implementation is approved I'll do a documentation pass

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -0,0 +1,99 @@
...
+            // This happens when UnityLogHandler is being used.
+            if (Log.IsHandlingLog)
+            {
+                m_ExistingLogHandler.LogException(exception, context);
+                return;
+            }
+
+            StackFrame stackFrame = GetLogCallerStackFrame();
+            object loggerContext = ResolveLogContext(context, stackFrame);
+            Log.GetLogger(loggerContext).Error(exception.ToString(), stackFrame.GetFileName(), stackFrame.GetMethod().Name, stackFrame.GetFileLineNumber());
```
jkeon, line 66: I feel like these three lines and lines 48 to 50 could be consolidated into one function that both `LogFormat` and `LogException` call?

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        private StackFrame GetLogCallerStackFrame()
+        {
+            return new StackFrame(4, true);
```
jkeon, line 71: Nice to have a comment for the magic number 4?

jkeon (approved):
Looks good, two minor comments/improvements

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: I like that idea. 
The `GetLogCallerStackFrame` number gets even more ambiguous/"magic" then but I think if I pass in the depth from here that'll be fine.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: Good call. I can account for all but one of that value. There's some invisible stack entry that doesn't show up in logs that has to do with Unity's logging system.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: Fixed. 
Forced me to figure out what exactly was happening at frame 2 :)
Also decided against the passed in skipFrames parameter. The comments are sufficient for now.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: Fixed. See response on other comment

jkeon (approved):

mbaker3:
Still open to feedback but merging and can revise in a new PR.

PR #60, opened 2022-04-28
Logging - Ensure early init, workaround Unity.Debug

mbaker3:
- Add call to `Log` during subsystem registration phase to make sure the listener gets initialized ASAP
 - Add workaround for surprise new logger that Unity has in the works. Currently `Unity.Entities@v0.17` has a `Unity.Debug` static class that proxies logger calls until the new logger is rolled out. Unity source Comment:
``` csharp
    // TODO: provide an implementation of Unity.Debug that does not rely on UnityEngine and modernizes this API
    // (for now it's just here for easier compatibility and fwd migration)
```

 - `UnityLogListener`'s initialization is not necessarily initialized ahead of Unity code. The listener is initialized when `Log` is first touched. This could result in missed log messages.
 - Originating caller is incorrectly determined when `Unity.Debug` is used to emit logs. (not to be confused with `UnityEngine.Debug`)

 - `UnityLogListener` touches `Log` during subsystem registration to make sure it is initialized ASAP.
 - `UnityLogListener` checks the `StackFrame` to see whether it is referencing a `Unity.Debug` call. If true, a new `StackFrame` on level deeper is used.

 - Potentially missed log messages at the start of application execution
 - Incorrect caller context listed when `Unity.Debug` is used.

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
@@ -23,6 +23,8 @@ namespace Anvil.Unity.Logging
         private static void Init()
         {
             s_Instance = null;
+            //Touch Log system to make sure this listener gets re-created ASAP.
+            Log.GetStaticLogger(typeof(UnityLogListener));
         }
 
         // The log handler in place before this listener initialized.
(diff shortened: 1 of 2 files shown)


tjvezina (commented):
I wondered if there was a more dynamic way to check and skip any frames in the stack that come from Unity (ex. FullName starts with "Unity." or something) but I imagine that would get expensive to do for every log - is it always exactly 4 frames for all of Unity's log options (debug, warning, error...)?

mbaker3:
> I wondered if there was a more dynamic way to check and skip any frames in the stack that come from Unity (ex. FullName starts with "Unity." or something) but I imagine that would get expensive to do for every log

I thought about that but it casts too wide of a net. We do still want call sources from within the `Unity.` and `UnityEngine.` name spaces. Just not this specific Debug class.

You could come up with some wildcard search like `Unity**.Debug` but I'm worried about accidentally picking up unintended classes now or in the future.

I think for now it's fine as a hard-coded special case. So far it's the only exception I've found but if the exceptions start to get out of hand then we can try to come up with a more sophisticated approach.

Performance wise, Logs are expensive, `StackFrame` with the extra file info flag definitely isn't cheap. A loop wouldn't be the end of the world. This code wouldn't ever be executed in a release build and I'm not so worried about the performance in debug builds...within reason.

> - is it always exactly 4 frames for all of Unity's log options (debug, warning, error...)?

Yes the flow is `Debug.Log*() -> UnityEngine.Logger.Log*() -> UnityEngine.ILogHandler.LogFormat()`.
The Log{severity} API is kept intact up to the call to `ILogHandler.LogFormat` which they each do themselves.

jkeon (approved):

mbaker3:
Merged but comments still welcome

PR #61, opened 2022-05-09
UnityLogListener - Fix exception when context resolves to null

mbaker3:
In some circumstances `stackframe.GetMethod()` returns `null`. We now guard against that scenario. 

The `UnityLogListener` will throw an exception when the targeted `StackFrame` can't determine the calling method.
There are a number of edge cases where this can happen but this was discovered when updating to Unity.Entities@0.50 from code that was throwing exceptions from within the entities package.

Further reading indicates that there are many scenarios where `GetMethod()` will return null:
 - https://stackoverflow.com/a/21564897/640196
 - https://github.com/dotnet/runtime/issues/31412

When this happens we can't resolve the context so we set the context to the private `UnknownContext` type. At runtime the name of this context just resolves to an empty string.

This exception:
```
Invalid exception thrown from custom Anvil.Unity.Logging.UnityLogListener.LogException(). Message: System.NullReferenceException: Object reference not set to an instance of an object
  at Anvil.Unity.Logging.UnityLogListener.SendToLogger (UnityEngine.Object context, Anvil.CSharp.Logging.LogLevel logLevel, System.String message) [0x0005b] in /Users/mbaker/Projects/twisted-dots/build/TwistedDOTS/Assets/Anvil/unity/anvil-unity-core/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs:96 
  at Anvil.Unity.Logging.UnityLogListener.LogException (System.Exception exception, UnityEngine.Object context) [0x0001b] in /Users/mbaker/Projects/twisted-dots/build/TwistedDOTS/Assets/Anvil/unity/anvil-unity-core/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs:74 
  at UnityEngine.Logger.LogException (System.Exception exception, UnityEngine.Object context) [0x0000b] in /Users/bokken/buildslave/unity/build/Runtime/Export/Logging/Logger.cs:132 
  at UnityEngine.Debug.CallOverridenDebugHandler (System.Exception exception, UnityEngine.Object obj) [0x0001e] in /Users/bokken/buildslave/unity/build/Runtime/Export/Debug/Debug.bindings.cs:262 : Null
UnityEngine.Debug:CallOverridenDebugHandler (System.Exception,UnityEngine.Object)
```

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Reflection;
 using Anvil.CSharp.Logging;
 using UnityEngine;
 using StackFrame = System.Diagnostics.StackFrame;
@@ -12,6 +13,11 @@ namespace Anvil.Unity.Logging
     [DefaultLogListener]
     public sealed class UnityLogListener : ILogListener, UnityEngine.ILogHandler
     {
+        /// <summary>
+        /// The context delivered when a context can't be resolved
+        /// </summary>
+        private sealed class UnknownContext { }
+
         private static UnityLogListener s_Instance;
 
         // With the Burst compiler, it's common to disable the Domain Reload step when entering play mode to increase iteration time.
@@ -65,7 +71,7 @@ namespace Anvil.Unity.Logging
                 return;
             }
 
-            SendToLogger(context, LogLevel.Error, exception.ToString());   
+            SendToLogger(context, LogLevel.Error, exception.ToString());
         }
 
         private void SendToLogger(UnityEngine.Object context, LogLevel logLevel, string message)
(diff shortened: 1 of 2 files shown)


Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -80,18 +86,21 @@ private void SendToLogger(UnityEngine.Object context, LogLevel logLevel, string
...
+            logger.AtLevel(logLevel, message, stackFrame.GetFileName(), stackFrame.GetMethod()?.Name, stackFrame.GetFileLineNumber());
         }
 
-        private object ResolveLogContext(UnityEngine.Object unityContext, StackFrame stackFrame)
+        private Type ResolveContextFromStack(StackFrame stackFrame)
         {
-            return unityContext == null ? stackFrame.GetMethod().ReflectedType : (object)unityContext;
+            MethodBase method = stackFrame.GetMethod();
+
+            return method != null ? method.ReflectedType : typeof(UnknownContext);
```
tjvezina, line 103: Looks like it used to use `Log.GetLogger(...)` in all cases before, but now if the context is null it uses `Log.GetStaticLogger(...)`, what's the difference here?

mbaker3: `Log.GetStaticLogger` vs `Log.GetLogger` both give you a `Log.Logger` instance but using different parameters. The idea is that most of the time you're in a non-static class and want to build a logger specific to that instance so you pass in `this`.

With static classes though when we build a logger we don't have an instance so we have to pass in the type.

Really, this just resolves to setting `Log.Logger.m_DerivedTypeName` with `insance.GetType().Name`(GetLogger) vs `type.Name`(GetStaticLogger)

---

In this specific case `GetStaticLogger` is used because we only have access to the calling type (at best) and not the calling instance.

tjvezina (approved):

PR #62, opened 2022-05-17
UnityLogListener - Fix logs from a burst compiled context

mbaker3:
Log messages emitted from bust compiled code are now caught by `UnityLogListener` and piped through the `Log` system.

#### Tag Alongs
**ExcludeAttribute** - Add Attribute which may be placed on methods to exclude them when attempting to resolve the context of a log message caller. This is used to support more robust caller resolution and allow specialized loggers to be less careful about internal stack depth before triggering the `UnityLogListener`.

`UnityLogListener` does not receive events emitted from Burst compiled code. This is because logs from burst compiled code do not go through the `Unity.unityLogger` which is where the `UnityLogListener` injects itself to observe logs.

As a result our log handlers would never get a chance to see/handle the messages.
This is unfortunate because the only way to emit logs is through `Debug.Log`
(and soon `BurstableLogger` but that's a different PR)

Luckily logs from burst compiled code do trigger `Application.logMessageReceivedThreaded` which the `UnityLogListener` now listens to.

This event triggers on whatever thread the log is emitted so the message must be marshalled back to the main thread. This is done by adding the message to a `ConcurrentQueue` for main thread processing. Pending log messages in the queue are processed as often as possible using the various `MonoBehaviour` life-cycle events (`OnUpdate`, `OnPreRender`, etc...). This does mean that the order of log messages between the Unity Editor Console and a `ILogHandler` (Ex: `FileLogHandler) may not be quite the same. However, developers shouldn't be expecting exact ordering when reading logs emitted from multi-threadded code anyway.

Unfortunately, there is currently no way to prevent the log message emitted by burst from appearing in the Editor console window ([Discussion](https://forum.unity.com/threads/is-it-possible-to-disable-debug-log-in-burst-compiled-code.1280819/)). To avoid doubling up logging messages `UnityLogListener` sets a flag when processing pending events so that it does not post the `Log` system decorated message to the console when the message comes back around.
Normal flow (simplified) is `Debug.Log -> UnityLogListener -> Log -> UnityLogHandler -> Deubg.Log -> UnityLogListener`.

The flow is a lot to take in but it has been documented in detail in `UnityLogListener.Application_logMessageReceivedThreaded`.

#### ExcludeAttribute
When resolving the caller before sending the message off to `Log` the logic now walks up the stack until a method without `Exclude` attribute is found.

Logs emitted from burst compiled code are now piped through the `Log` system and decorated for all handlers except `UnityLogHandler`

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
@@ -234,7 +234,7 @@ namespace Anvil.Unity.Logging
             //       - UnityLogHandler.HandleLog
             //        - Debug.Log
             //         - UnityLogListener.LogFormat - Do nothing because m_IsHandlingBurstedLog = true
-            //  - Someting external (BurstCompilerService.Log?) emits the log to the Unity console
+            //  - Something external (BurstCompilerService.Log?) emits the log to the Unity console
             //      
             // From a Log.Logger instance
             //  - Log.Logger.Debug
@@ -247,9 +247,6 @@ namespace Anvil.Unity.Logging
             //         - UnityLogListener.Application_logMessageReceivedThreaded - Do nothing because Log.IsHandling == true
             //        - m_ExistingLogHandler emits log to the Unity console
             m_PendingLogs.Enqueue(new LogMessage(condition, LogTypeToLogLevel(type), true));
-            //m_IsHandlingBurstedLog = true;
-            //SendToLogger(null, LogTypeToLogLevel(type), condition);
-            //m_IsHandlingBurstedLog = false;
         }
 
         private void PendingLogPump_OnProcessPendingLogs()
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
@@ -3,6 +3,7 @@ using System.Reflection;
 using Anvil.CSharp.Logging;
 using UnityEngine;
 using StackFrame = System.Diagnostics.StackFrame;
+using System.Collections.Concurrent;
 
 namespace Anvil.Unity.Logging
 {
@@ -17,6 +18,30 @@ namespace Anvil.Unity.Logging
         /// The context delivered when a context can't be resolved
         /// </summary>
         private sealed class UnknownContext { }
+
+        private readonly struct LogMessage
+        {
+            public readonly string Message;
+            public readonly LogLevel LogLevel;
+            public readonly bool IsFromBurstedContext;
+
+            public LogMessage(string message, LogLevel logLevel, bool isFromBurstedContext)
+            {
+                Message = message;
+                LogLevel = logLevel;
+                IsFromBurstedContext = isFromBurstedContext;
+            }
+        }
+
+        private class LateUpdatePump : MonoBehaviour
+        {
+            public event Action OnLateUpdate;
+
+            private void LateUpdate()
+            {
+                OnLateUpdate?.Invoke();
+            }
+        }
 
         private static UnityLogListener s_Instance;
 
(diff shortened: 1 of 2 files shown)


Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -13,11 +15,96 @@ namespace Anvil.Unity.Logging
...
+                IsFromBurstedContext = isFromBurstedContext;
+            }
+        }
+
+        // Pumps events off the main thread to indicate that pending logs
+        // should be processed.
+        // We want to process the queue as often as possible (within reason) so we can handle
+        // the logs as close to the time they were emitted as possible.
+        [DefaultExecutionOrder(int.MaxValue)]   // Execute last in each phase
+        private sealed class PendingLogPump : MonoBehaviour
```
mbaker3, line 49: A custom pump was created because the Logging project doesn't reference the greater anvil-unity-core framework and lives within it as a DLL.

This would be a terribly inefficient use case for `UpdateHandle` anyway!

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -13,11 +15,96 @@ namespace Anvil.Unity.Logging
...
+
+            private void OnPostRender()
+            {
+                OnProcessPendingLogs?.Invoke();
+            }
+
+            private void OnGUI()
+            {
+                OnProcessPendingLogs?.Invoke();
+            }
```
jkeon, line 87: I'm not sure if we should have this one. 

This one can be called 100 times a frame to update the immediate mode GUI. 

The next step is EndOfFrame and we're processing there again.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -67,40 +191,134 @@ public void LogException(Exception exception, UnityEngine.Object context)
...
+            //     - Debug.Log
+            //      - UnityLogListener.LogFormat
+            //       - m_ExistingLogHandler.LogFormat - because Log.IsHandlingLog == true
+            //        - Application.logMessageReceived(+Threaded)
+            //         - UnityLogListener.Application_logMessageReceivedThreaded - Do nothing because Log.IsHandling == true
+            //        - m_ExistingLogHandler emits log to the Unity console
+            m_PendingLogs.Enqueue(new LogMessage(condition, LogTypeToLogLevel(type), true));
+            //m_IsHandlingBurstedLog = true;
+            //SendToLogger(null, LogTypeToLogLevel(type), condition);
+            //m_IsHandlingBurstedLog = false;
```
jkeon, line 257: Should these be removed?

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -67,40 +191,134 @@ public void LogException(Exception exception, UnityEngine.Object context)
...
+            logger.AtLevel(logLevel, message, callerFrame.GetFileName(), callerMethod?.Name, callerFrame.GetFileLineNumber());
+        }
+
+        private (StackFrame callerFrame, MethodBase callerMethod) ResolveCaller(int skipFrames)
+        {
+            StackFrame callerFrame;
+            MethodBase callerMethod;
+            // Walk up the stack until we're at a caller that isn't Unity's proxy or a method that we 
+            // want to skip (Ex: Logger objects).
+            do
```
jkeon, line 297: First i've ever seen this in the wild.

jkeon (changes requested):
Looks good, couple little things.

Also:

> However, developers shouldn't be expecting exact ordering when reading logs emitted from multi-threadded code anyway.

Agreed, I wonder if there is value or if it is possible for you to put a prefix on any logs that come from off the main thread? 

Ex. [OffThread]

Might just make it a nice reminder that your order is going to be weird and you don't get down a debugging rabbit hole

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: sure, makes sense. I forgot how often this is hammered. Fixed

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: It's an elusive one!

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: Yeep, fixed

mbaker3:
> > However, developers shouldn't be expecting exact ordering when reading logs emitted from multi-threadded code anyway.
> 
> Agreed, I wonder if there is value or if it is possible for you to put a prefix on any logs that come from off the main thread?
> 
> Ex. [OffThread]
> 
> Might just make it a nice reminder that your order is going to be weird and you don't get down a debugging rabbit hole

Good call. log messages from `BurstableLogger` are now decorated with `(Thread: {threadIndex})` if not on the main thread. (https://github.com/decline-cookies/anvil-unity-dots/pull/30/commits/f11a7e78f529acf78aefdd44cb4634c80434f40d)

jkeon (approved):

PR #64, opened 2022-06-27
Hotfix suppress LINQ warnings

mbaker3:
When [NetFabric.Hyperlinq.Analyzer](https://github.com/NetFabric/NetFabric.Hyperlinq.Analyzer) is installed it emits warnings about the use of `.Single()` because most of the time `.First()` is more appropriate.

Specific Error: https://github.com/NetFabric/NetFabric.Hyperlinq.Analyzer/blob/master/docs/reference/HLQ005_AvoidSingle.md

Warnings are emitted for the use of `.Single()` in `LogProjectUtils`

The warnings are suppressed.

In this case the use of `Single` is appropriate since we're asserting that there is only one of the queried value at the same time as the query.

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Editor/Logging/LogProjectUtils.cs
+++ b/Scripts/Editor/Logging/LogProjectUtils.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Xml.Linq;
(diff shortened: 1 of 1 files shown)


jkeon (approved):
Looks fine to me but I'll let @tjvezina weigh in since he was part of this from the beginning.

tjvezina (approved):
Huh, weird warning

mbaker3:
> Huh, weird warning

Yea it's a bit heavy handed imo but the potential performance impacts are valid vs `.First()`

PR #66, opened 2022-08-02
Log Updates

mbaker3:
Updates to support changes from https://github.com/decline-cookies/anvil-csharp-core/pull/91

 - Uses of Logger are not compatible with latest changes to underlying Log types.
 - OneTime log messages do not emit after the first play in the editor when domain reload is disabled.

 - Update references to support changes from https://github.com/decline-cookies/anvil-csharp-core/pull/91
 - Create `StaticStateResetHelper` in the Unity Logging dll to allow the state of the one time logs to be reset when domain reload is disabled.

 - None

 - https://github.com/decline-cookies/anvil-csharp-core/pull/91

 - [ ] Yes 
 - [x] No


jkeon (approved):
Looks good

tjvezina (approved):

PR #68, opened 2022-08-02
Add ScreenUtil

mbaker3:
Add a collection of utilities to help deal with screens and resolutions.

 - None

Introduce methods to:
 - Provide normalized values between the edge of the screen and an arbitrary distance.
 - Normalize a scale value from one DPI to another.

Includes additional `Unity.Mathematics` dependency

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Core/Util/ScreenUtil.cs
+++ b/Scripts/Runtime/Core/Util/ScreenUtil.cs
@@ -1,0 +1,126 @@
+using Anvil.CSharp.Logging;
+using Unity.Mathematics;
+using UnityEngine;
+
+namespace Anvil.Unity.Core
+{
+    /// <summary>
+    /// A collection of utilities for dealing with display screen resolutions
+    /// </summary>
+    public static class ScreenUtil
+    {
+        /// <summary>
+        /// The assumed base DPI for all edge border values provided to utility methods below.
+        /// This is used to normalize values against the current screen's DPI.
+        /// </summary>
+        public const float BASE_DPI = 102f;
+
+        /// <summary>
+        /// Normalizes an input position to the edges of <see cref="Screen"/> within the borders provided where bottom
+        /// left is (-1,-1) and top right is (1,1).
+        /// </summary>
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Core/Util/ScreenUtil.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// <summary>
+    /// A collection of utilities for dealing with display screen resolutions
+    /// </summary>
+    public static class ScreenUtil
+    {
+        /// <summary>
+        /// The assumed base DPI for all edge border values provided to utility methods below.
+        /// This is used to normalize values against the current screen's DPI.
+        /// </summary>
+        public const float BASE_DPI = 102f;
```
jkeon, line 16: Why 102? Where did we get that from?

jkeon (approved):

Scripts/Runtime/Core/Util/ScreenUtil.cs
mbaker3: This what `Screen.dpi` reports on my [Dell U3011](https://www1.la.dell.com/ky/en/corp/peripherals/monitor-dell-u3011/pd.aspx?refid=monitor-dell-u3011&s=corp)

On my M1 Macbook Pro display it reports `255`.

Interestingly, to get the correct DPI in the editor you have to wait a bit. Logging Screen.dpi in the application class always reported my laptop's DPI. Waiting until the `Awake()` of a view's code provided the correct value for the monitor being rendered on.

At the end of the day we need some sort of reasonable base. I'm open to low effort suggestions but I think the only solution would be to have a comprehensive set of tools that adjust author time values based on the developer's DPI at author time. So setting size values on a retina display would appear as differently set values on a non-retina display.

jkeon: Hmmn ok. I did a bit of research and it's all over the map. I guess keeping this is fine until we notice issues and have a better base.

Issue #47, opened 2021-03-27
Support Higher Levels of AssemblyStripping


mbaker3:
Currently Anvil does not support higher levels of Assembly Stripping than `Low` (Unity 2021.1.0).

When the compiler is set to higher levels of stripping the constructors for some classes are stripped out of the build. Constructors are stripped because they do not have a hard reference anywhere. This is common for classes that are instantiated through `Activator.CreateInstance` to incorporate a specific implementation into a more generic system (Ex: Update sources for `Anvil.CSharp.DelayedExecution`.

Some issues will require fixes in [anvil-csharp-core](https://github.com/scratch-games/anvil-csharp-core). This is tracked via: https://github.com/scratch-games/anvil-csharp-core/issues/49

See the following links for more causes of stripping related issues and potential solutions.
- https://docs.unity3d.com/Manual/ScriptingRestrictions.html
- https://docs.unity3d.com/Manual/ManagedCodeStripping.html
   - `Link.xml` was used to resolve some stripping issues as part of #46. It may not be suitable for all situations

# Examples
- All subclasses of `AbstractUnityUpdateSource`. For example, `UnityFixedUpdateSource`
   - The simple fix is to mark all `AbstractUnityUpdateSource` subclasses with the [UnityEngine.Scripting.PreserveAttribute](https://docs.unity3d.com/2021.1/Documentation/ScriptReference/Scripting.PreserveAttribute.html) but this isn't very maintainable and keeps the classes even when they are not being used.
- TODO: Add more as they are encountered

# To Reproduce
1. Add anvil to a Unity project
2. Make use of one of the classes mentioned in the examples above
3. Set project settings to `Managed Stripping Level: High`
4. Export and run a build
5. Notice `ExecutionEngineException` thrown.

mbaker3:
Oh, I just put up a PR for this!
#74

Issue #73, opened 2022-08-17
Select a Unit Test Framework


mbaker3:
Related to: https://github.com/decline-cookies/anvil-csharp-core/issues/106, https://github.com/decline-cookies/anvil-unity-dots/issues/50

Anvil should really have Unit tests. There are already some issues defined to write tests. We just need to select a unit test framework.

Keep in mind that this is a pure Unity library so should be easy to execute from Unity.

PR #74, opened 2022-08-22
Improve Stripping Support

mbaker3:
These changes allow Anvil to support the highest level of bytecode stripping built into Unity and makes the existing stripping defences more targeted to the exact types that require it. 

When building a project that's using Anvil with [Managed Code Stripping](https://docs.unity3d.com/2021.3/Documentation/Manual/ManagedCodeStripping.html) set to `Low` or higher the build will throw exceptions.
 - All of the delayed execution update sources (Ex: `UnityFixedUpdateSource`) do not have default constructors
 - `UnityLogHandler` is not found as the default log handler.

#### Fix Update Sources
Each concrete implementation of an update source now has a default constructor defined decorated with `[RequiredMember]` .
https://docs.unity3d.com/2021.3/Documentation/ScriptReference/Scripting.RequiredMemberAttribute.html

This prevents the default constructor from being stripped when the update source is referenced in the project.

**Note:** I did attempt to use [RequireDerivedAttribute](https://docs.unity3d.com/2021.3/Documentation/ScriptReference/Scripting.RequireDerivedAttribute.html) but it didn't drag along the default constructor.

#### Fix `UnityLogHandler`
This is an oversight from #70 where `link.xml` wasn't updated to reflect the new assembly name.
While fixing the issue I decided to migrate away from using `link.xml` and specifically add `[Preserve]` attributes to the log listener and log handler implementations.

The advantage to this approach:
 - Other unused logging types may be stripped rather than all being included even if not required.
 - Not affected by renaming assemblies

**Note:** Unfortunately we couldn't use [RequireImplementors](https://docs.unity3d.com/2021.3/Documentation/ScriptReference/Scripting.RequireImplementorsAttribute.html) because these attributes are Unity specific and the interfaces are defined in `anvil-csharp-core`.

**Resolves:** #47

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/DelayedExecution/UpdateSources/UnityFixedUpdateSource.cs
+++ b/Scripts/Runtime/DelayedExecution/UpdateSources/UnityFixedUpdateSource.cs
@@ -1,4 +1,6 @@
-﻿namespace Anvil.Unity.DelayedExecution
+﻿using UnityEngine.Scripting;
+
+namespace Anvil.Unity.DelayedExecution
 {
     /// <summary>
     /// A specific <see cref="AbstractUnityUpdateSource"/> that dispatches
@@ -8,5 +10,8 @@
     /// </summary>
     public class UnityFixedUpdateSource : AbstractUnityUpdateSource<UnityFixedUpdateSourceMonoBehaviour>
     {
+        // Required to prevent constructor from being stripped when this type is referenced.
+        [RequiredMember]
+        public UnityFixedUpdateSource() : base() { }
     }
 }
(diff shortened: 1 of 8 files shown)


tjvezina (approved):
The changes make sense to me, but what do empty constructors have to do with build stripping anyway?

How do you avoid future code additions from causing more exceptions? I'm guessing your project will have build stripping enabled, so you'd find out when a build fails?

mbaker3:
> The changes make sense to me, but what do empty constructors have to do with build stripping anyway?

Stripping can be pretty aggressive. If it determines that a type is never instantiated by its default constructor it will strip that constructor away. It'll also strip out unused interfaces, methods, and attributes.

I found out that default constructors were being stripped off the update sources when I enabled stripping on my project.

In the future, the work done for #72 or a specific test after completing #73 will catch these issues at merge time.

jkeon (approved):

PR #76, opened 2022-09-06
ID provider static domain reloading.

jkeon:
The `ID` static class in https://github.com/decline-cookies/anvil-csharp-core/pull/109 could preserve unwanted state when running in the editor with domain reloading turned off.

ID Providers could stick around after a soft-reload in the app.

ID providers are cleaned up.

- None

 - [Anvil CSharp #109](https://github.com/decline-cookies/anvil-csharp-core/pull/109)

 - [x] Yes
 - [ ] No

### DIFF
--- a/Scripts/Runtime/Data/ID/ID.cs
+++ b/Scripts/Runtime/Data/ID/ID.cs
@@ -1,0 +1,17 @@
+using UnityEngine;
+
+namespace Anvil.Unity.Data
+{
+    /// <summary>
+    /// Resets the <see cref="CSharp.Data.ID"/> static class on load.
+    /// This allows for Unity's domain reloading to be disabled which speeds up Burst compiler development.
+    /// </summary>
+    internal static class ID
+    {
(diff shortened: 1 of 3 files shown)


Scripts/Runtime/Data/ID/ID.cs
```diff
@@ -0,0 +1,17 @@
+using UnityEngine;
+
+namespace Anvil.Unity.Data
+{
+    /// <summary>
+    /// Resets the <see cref="CSharp.Data.ID"/> static class on load.
+    /// This allows for Unity's domain reloading to be disabled which speeds up Burst compiler development.
+    /// </summary>
+    internal static class ID
```
mbaker3, line 9: Can we call this `StaticStateResetHelper` to be consistent with the one in Logging?

jkeon:
Closing as a change in https://github.com/decline-cookies/anvil-csharp-core/pull/109 made this not necessary.

PR #77, opened 2022-09-22
Unity TraceListener

tjvezina:
Added a `TraceListener` for Unity, to handle `System.Diagnostics.Debug.Assert()` calls and others.

Note: `System.Diagnostics.Debug...` and `System.Diagnostics.Trace...` functions map to the same listener, the difference is that `Debug` functions only execute in debug builds (i.e. when the `DEBUG` symbol is defined, which Unity does), while `Trace` works in all builds.

Any calls under `System.Diagnostics.Debug/Trace` have no effect in Unity, as there's no listener mapping its called to Unity functions. I.e. `System.Diagnostics.Debug.Assert(...)` calls are ignored, even on failure.

All `System.Diagnostics.Debug/Trace` calls work as expected, forwarding messages to Unity's console or throwing exceptions.

Non-Unity code (ex. anvil-csharp-core) that uses `System.Diagnostics` functions for writing or assertions are ignored in Unity projects. This listeners ensures they will now work as expected.

 - None

 - [ ] Yes
 - [X] No

### DIFF
--- a/Scripts/Runtime/Debug/UnityTraceListener.cs
+++ b/Scripts/Runtime/Debug/UnityTraceListener.cs
@@ -1,0 +1,31 @@
+using System;
+using UnityEngine;
+
+namespace Anvil.Unity.Debugging
+{
+    /// <summary>
+    /// A custom <see cref="System.Diagnostics.TraceListener"> for Unity, to allow the use of <see cref="System.Diagnostics"> functions.
+    /// By default, all of the .NET Debug/Trace functions have no effect in Unity, without a listener manually forwarding calls.
+    /// 
+    /// The most common and important functions enabled by this listener are:
+    /// - <see cref="System.Diagnostics.Debug.WriteLine"> (forwarded to UnityEngine.Debug.Log)
+    /// - <see cref="System.Diagnostics.Debug.Assert"> (forwarded to a thrown exception)
+    /// 
+    /// Note: Both <see cref="System.Diagnostics.Debug"> and <see cref="System.Diagnostics.Trace"> functions use the same
+    /// listener, the difference is that Debug functions become no-ops in release builds, while Trace always works.
+    /// </summary>
+    public class UnityTraceListener : System.Diagnostics.TraceListener
+    {
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Init()
+        {
+            System.Diagnostics.Trace.Listeners.Add(new UnityTraceListener());
+        }
+
+        public override void Write(string message) { UnityEngine.Debug.Log(message); }
+        public override void WriteLine(string message) { UnityEngine.Debug.Log(message); }
+
+        public override void Fail(string message) { throw new Exception(message); }
+        public override void Fail(string message, string details) { throw new Exception($"{message} {details}"); }
+    }
+}
--- a/Scripts/Runtime/Debug.meta
+++ b/Scripts/Runtime/Debug.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 3deaa595a3b50bd4985008fd95b5b3cc
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Scripts/Runtime/Debug/UnityTraceListener.cs.meta
+++ b/Scripts/Runtime/Debug/UnityTraceListener.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: fd35388da8ade1941abc374f543e5d9a
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 


Scripts/Runtime/Debug/UnityTraceListener.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+using UnityEngine;
+
+namespace Anvil.Unity.Debugging
```
tjvezina, line 4: I'm open to suggestions on the folder/namespace naming. It just can't be `Anvil.Unity.Debug` for naming conflict reasons.

Scripts/Runtime/Debug/UnityTraceListener.cs
```diff
@@ -0,0 +1,31 @@
...
+        private static void Init()
+        {
+            System.Diagnostics.Trace.Listeners.Add(new UnityTraceListener());
+        }
+
+        public override void Write(string message) { UnityEngine.Debug.Log(message); }
+        public override void WriteLine(string message) { UnityEngine.Debug.Log(message); }
+
+        public override void Fail(string message) { throw new Exception(message); }
+        public override void Fail(string message, string details) { throw new Exception($"{message} {details}"); }
```
tjvezina, line 29: I asked this question in https://github.com/decline-cookies/anvil-unity-core/issues/30 as well, but - should this throw an exception, or just `Debug.LogError(message)`?

tjvezina, line 29: These are technically public functions, but never meant to be accessible to users, so I wasn't sure if they needed header comments.

Scripts/Runtime/Debug/UnityTraceListener.cs
jkeon: I think this works, the `csharp` library has a `Anvil.CSharp.Debugging` namespace for `AssertUtil` so i think we're being consistent.

Scripts/Runtime/Debug/UnityTraceListener.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// 
+    /// Note: Both <see cref="System.Diagnostics.Debug"> and <see cref="System.Diagnostics.Trace"> functions use the same
+    /// listener, the difference is that Debug functions become no-ops in release builds, while Trace always works.
+    /// </summary>
+    public class UnityTraceListener : System.Diagnostics.TraceListener
+    {
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Init()
+        {
+            System.Diagnostics.Trace.Listeners.Add(new UnityTraceListener());
```
jkeon, line 22: Do we get duplicates in here with Domain Reloading turned off and multiple compiles of the code? 

See: https://docs.unity3d.com/2021.3/Documentation/Manual/DomainReloading.html

My gut would say that this looks like we're adding a new instance to a static list each time which will lead to duplicates but I'm not sure how the Listeners work under the hood.

Scripts/Runtime/Debug/UnityTraceListener.cs
jkeon: I think exception. I'll link my full response here: https://github.com/decline-cookies/anvil-unity-core/issues/30#issuecomment-1254938413

jkeon: Can we just `/// <inheritdoc cref="TraceListener.X"/>`?

jkeon (changes requested):
Looks great, not really "request changes" but just some open questions.

Scripts/Runtime/Debug/UnityTraceListener.cs
tjvezina: Hmm I'm not sure, I'll test it out and add a static instance if necessary.

Scripts/Runtime/Debug/UnityTraceListener.cs
tjvezina: I can, but that class too is not meant to be user-accessible, and so has no docs to inherit. :P Should I add those anyway?

jkeon: Ah gotcha. In that case maybe just make this class `internal` 

We never need to interact with it and the static initializer handles creating and registering it.

Scripts/Runtime/Debug/UnityTraceListener.cs
tjvezina: I'm getting confusing results - with or without "Enter Play Mode Settings > Reload Domain" enabled in editor settings, this static function is only ever called once for me, unless I'm not triggering domain reloading correctly.

I.e. if I throw a log in that static `Init()` and start play mode, I see it log once - then if I make a code change and go back to Unity, it reloads scripts but I don't see that log again, and there's no duplicate listener.

I considered adding a `static s_Instance` anyway, but based on the above link, domain reloading would reset static fields anyway? I'm unsure how to handle this case - there doesn't seem to be anything going wrong, but I might be missing a use case. I will continue researching, but I'm open to suggestions.

tjvezina: Ahhh my confusion comes from the editor settings - I thought Reload Domain was disabled by default because it's unchecked, but if "Enter Play Mode Options" is also unchecked, it defaults to enabled, so I wasn't actually testing the disabled state.

Would this be sufficient?
```cs
if (s_Instance == null)
{
    s_Instance = new UnityTraceListener();
    System.Diagnostics.Trace.Listeners.Add(s_Instance);
}
```

jkeon (dismissed):
Looks great!

tjvezina:
FYI when an assert fails, the log context is broken/missing:
```
(|:0) System.Exception: Exception of type 'System.Exception' was thrown.
```
but that's unrelated to this PR, that's also true if you simply `throw new Exception(...)` from a script.

Scripts/Runtime/Debug/UnityTraceListener.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            s_Instance = new UnityTraceListener();
+            System.Diagnostics.Trace.Listeners.Add(s_Instance);
+        }
+
+        private UnityTraceListener() { }
+
+        public override void Write(string message) { UnityEngine.Debug.Log(message); }
+        public override void WriteLine(string message) { UnityEngine.Debug.Log(message); }
```
mbaker3, line 37: We should probably use Logger here. Something like:
`Log.GetStaticLogger(typeof(Test), "[ASSERT]")`

Although we caching the logger instance is likely our best bet.

If we use logger we can also include all of the "correct" context for the call site. Check out `UnityLogListener.ResolveCaller` for an example of this.

mbaker3 (approved):
Looks good. Left a comment about improving log context. We can defer to a future feature though. This PR without the changes still adds value.
Up to you @tjvezina.

Scripts/Runtime/Debug/UnityTraceListener.cs
tjvezina: I like the idea, but I'm not sure how to implement - where can I tell it to skip additional stack frames so the context falls on whoever called `System.Diagnostics.<something>` in the first place?

Would this replace the thrown exceptions too, or just the `Write`/`WriteLine` calls? I'm not sure where the `"[ASSERT]"` prefix comes in, and I assume `typeof(Test)` is a placeholder, but I don't know what context to actually use. If I try `Log.GetStaticLogger(typeof(UnityTraceListener)).Debug(message)` the output is the same as plain `Debug.Log(...)` after it gets picked up by Logger.

mbaker3: > Where can I tell it to skip additional stack frames so the context falls on whoever called System.Diagnostics.<something> in the first place?

My thought is that you either copy the functionality of `UnityLogListener.ResolveCaller` and hard code the number of stack steps you'll be up from the original call site or break `UnityLogListener.ResolveCaller` out into its own utility method making the functionality available to all. I can't remember whether there are log listener details baked into that method but I can investigate late next week if needed.

You would then use the information from `ResolveCaller` to populate the optional arguments in `Logger.Debug` the way that we do in `UnityLogListener`

`UnityLogListener.SendToLogger` might be better to review to see how these pieces come together to build the context from stack frames. Again though, you'll have to either copy or break out this functionality. `UnityLogListener` itself isn't going to help you here and is really just the last line of defence in trying to pipe information through `Log`


> Would this replace the thrown exceptions too

No, I think the exception is preferred. It will should include the stack anyway.

> Assume typeof(Test) is a placeholder

oops yes `Test` should have been replaced before I commented. On second thought though you should be able to build a logger using the type from the original call site `UnityLogListener.ResolveContextFromMethod` does this.


> I'm not sure where the "[ASSERT]" prefix comes in

I figured it would be helpful to prefix the log messages with that but it's not necessary if the listener will be picking up non-assert messages.

tjvezina: Thanks for the details - I'll give this a shot sometime before the PR is closed.

I've never seen `System.Diagnostics.Debug.WriteLine` used, I've always used `Console.WriteLine`, but a quick search tells me I've wrongly carried that practice over from C# console apps, and `Debug.WriteLine` is correct for debugging.

I don't see any `Console.WriteLine` uses in Anvil, but we do have a `ConsoleLogHandler` (presumably for console apps) - should there also be handlers for `System.Diagnostics.Debug` and `.Trace`? On the one hand, ideally all logs (within Anvil) use Anvil's Logger directly, but on the other hand Logger doesn't differentiate between `Debug` and `Trace` logging, and it doesn't make sense to use either (or both) for all logs in the app. 🤔

tjvezina:
Pushed some changes, mostly to the `UnityLogListener`, to fix some issues with resolving log context.

---

First, added `[UnityLogListener.Exclude]` attributes to the trace listener callbacks, so that `System.Diagnostics.Debug.WriteLine(...)` calls caught by the trace listener and forwarded to Unity correctly have the context of the original log. I.e. previously, logs looked like this (see first screenshot at top of PR):
```
(UnityTraceListener|WriteLine:26) Some log
```
and now they look like this:
```
(MyScript|SomeFunc:42) Some log
```

---

Second, fixed log context resolution for exceptions - the problem was, when an uncaught exception is picked up by Unity and sent to the log system, it now has a new useless call stack (basically just `UnityLogListener.LogException` alone). `LogException` now uses `new StackTrace(exception)` to pull the original stack trace for the code that caused the exception, and passes that up through the logging system instead.

So, where exceptions handled by `UnityLogListener` previously looked like this:
```
(|:0) System.Exception: Exception of type 'System.Exception' was thrown.
```
they now correctly look like this:
```
(MyScript|SomeFunc:42) System.Exception: Exception of type 'System.Exception' was thrown.
```

---

And finally, in case of exceptions/logs sent from a DLL (ex. Anvil's logging libraries themselves), there is no file name or number to resolve, so for example the exception in `UnityLogListener`'s constructor would look like this:
```
(|.ctor:0) System.Exception: There can only be one instance of the UnityLogListener and it has already been initialized.
```
Now, if the stack frame's file name is null, it falls back on the method's owning type:
```
(UnityLogListener|.ctor:0) System.Exception: There can only be one instance of the UnityLogListener and it has already been initialized.
```

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -103,6 +106,20 @@ private void OnDestroy()
...
+            "System.Diagnostics.Debug",
+            "System.Diagnostics.Trace",
+            "System.Diagnostics.TraceInternal",
+            "UnityEngine.Debug",
+            "UnityEngine.Logger",
+            "UnityEngine.Assertions",
+            // Check for Unity's new stubbed out logging class that currently just proxies the existing logging system.
+            // This was found in the v0.17 of the Unity.Entities package
+            "Unity.Debug",
+        };
```
tjvezina, line 121: As I ran into multiple flows used by Unity to send logs through its log handlers, I opted for a more flexible, if slightly less performant approach to resolving log contexts - rather that assume the flow will always be a fixed sequence of functions (especially in future Unity versions), I explicitly skip the types/namespaces listed here, as well as anything with Anvil's `Exclude` attribute.

The difference is a couple of `new StackFrame(...)` calls per log, which should not make a noticeable difference (that's something like 0.01ms), but if it's still a performance concern I can go back to making assumptions about Unity's logging call stack to improve it.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -145,7 +162,10 @@ private static void Init()
 
         public UnityLogListener()
         {
-            Debug.Assert(s_Instance == null, $"There can only be one instance of the {nameof(UnityLogListener)} and it has already been initialized.");
+            if (s_Instance != null)
+            {
+                throw new Exception($"There can only be one instance of the {nameof(UnityLogListener)} and it has already been initialized.");
+            }
```
tjvezina, line 168: I realized that all of the `Debug.Assert` checks in both the C# or Unity logging libraries are useless, because the DLL's are not debug builds.

In this specific case, I changed the assert to an exception because it's still needed in release builds too (in case an Anvil user tries to make their own listener), but how should the rest of the asserts be handled? Should the logging DLL's be changed to debug builds in the repos? Am I just too lazy, and I should've been switching to debug builds when making changes to those libraries? 😛

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -179,10 +200,19 @@ public void LogFormat(LogType logType, UnityEngine.Object context, string format
...
             }
 
-            SendToLogger(context, LogTypeToLogLevel(logType), string.Format(format, args));
+            // Make UnityEngine.Debug.Assert failures throw an exception instead of just logging an error
+            StackTrace stackTrace = new StackTrace(fNeedFileInfo: true);
+            MethodBase assertMethod = stackTrace.GetFrame(2)?.GetMethod(); // Skip this function and UnityEngine.Logger
+            if (assertMethod != null && assertMethod.Name == "Assert" && assertMethod.DeclaringType.FullName == "UnityEngine.Debug")
+            {
+                throw new Exception(string.Format(format, args));
+            }
```
tjvezina, line 209: This is open to debate, but given our decision that failed assertions should throw exceptions, I decided to intercept error logs from `UnityEngine.Debug.Assert` and convert them to exceptions. Let me know what you think.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
jkeon: I'm not sure, I'll leave this one for @mbaker3 to chime in on.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
jkeon: I think this makes sense to me. If something bad happens that  a Debug.Assert gets thrown, I really want to know about it and I think it should halt execution.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -267,47 +297,48 @@ private void ProcessPendingLogs()
...
-                    // This was found in the v0.17 of the Unity.Entities package
-                    || callerMethod.ReflectedType.FullName == "Unity.Debug"
-                    )
-                );
-
-            return (callerFrame, callerMethod);
+                MethodBase method = frame.GetMethod();
+
+                ExcludeAttribute excludeAttribute = method.GetCustomAttribute<ExcludeAttribute>(inherit: true);
+                if (excludeAttribute != null) continue;
```
jkeon, line 321: Minor but we like to keep the squigglies. 
```suggestion
                if (excludeAttribute != null) 
                {
                    continue;
                }
```

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -267,47 +297,48 @@ private void ProcessPendingLogs()
...
+                
+                string typeName = method.DeclaringType.FullName;
+                if (SKIPPED_STACK_FRAME_TYPES.Any(name => typeName.StartsWith(name))) continue;
```
jkeon, line 324: Minor squigglies comment again and relating up to earlier comment about making `SKIPPED_STACK_FRAME_TYPES` a `HashSet` instead.

```suggestion
                if (SKIPPED_STACK_FRAME_TYPES.Any(name => typeName.StartsWith(name))) 
                {
                    continue;
                }
```

jkeon (changes requested):
Looks good, some minor comments. 

Awesome that you solved those issues where we weren't seeing the context of the log!

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
tjvezina: Do you mean so I can say `SKIPPED_STACK_FRAME_TYPES.Contains(typeName)`? `Any` will short circuit when it finds a match, but I guess it always checks the whole array once when it finds the original context...

I think that works - I originally had `.StartsWith(typeName)` but I think I ended up with only full type names in this list (ex. initially I ignored anything under `System.Diagnostics.*`) so I could change it to a HashSet.Contains()

tjvezina: I lied, `.StartsWith(...)` was important because "UnityEngine.Assertions" is a namespace, so I had to specify "UnityEngine.Assertions.Assert" (and "System.Diagnostics.TraceListener") - if we catch any other internal logging types showing up as the log context in the future, they'll just have to be added to this list.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -267,47 +299,52 @@ private void ProcessPendingLogs()
...
-            // Skip 4 frames to get to the original caller
-            // 0 - Here
-            // 1 - LogException or LogFormat
-            // 2 - UnityEngine.Logger.Log(+Warning, +Error, ...), Assert
-            // 3 - UnityEngine.Debug.Log(+Warning, +Error, ...), Assert
-            // 4 - Caller of Debug.Log(+Warning, +Error, ...), Assert
-            (StackFrame callerFrame, MethodBase callerMethod) = ResolveCaller(4);
+            (StackFrame callerFrame, MethodBase callerMethod) = ResolveCaller(stackTrace);
+
+            string callerPath = callerFrame?.GetFileName() ?? callerMethod?.ReflectedType.FullName;
```
tjvezina, line 306: I changed this fallback to `ReflectedType.FullName` instead of just `Name` because the type name alone in library code was proving insufficient context.

However, I'm only seeing the namespace in the console, because `UnityLogHandler` is expecting an absolute file path and running `Path.GetFileNameWithoutExtension(...)` on it, which trims the "extension" (type name) off the end :P

I could move that up, but this leads to a greater discussion - should `callerPath` be more generic, like `callerName` or something? Even in the default case, would the `method.ReflectedType.Name` be more useful than the `frame.GetFileName()` even though those are typically the same?

Scripts/Runtime/Debug/UnityTraceListener.cs
mbaker3: Anything `*LogHandler` is for log outputting so `ConsoleLogHandler` is in fact for piping logger messages into `Console`. The `*LogListeners` are what capture logs made to other logging systems.

I suppose a `DebugLogHandler` is reasonable to have but I've never felt the need for one. Like you, I've always just carried over that `Console.WriteLine` was for debugging in console apps. 🤷 

Writing Log Listeners for Debug and Trace some day makes sense but isn't urgent until someone consumes a library that makes use of them and implementing the listener should be trivial with the hooks I'm seeing.

I'd say hold off on implementing any `Debug` or `Trace` log handlers for now.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: uhhh, let me investigate this one a bit. I think there is some compiler magic we can use with Unity to handle this.
With a normal C# project you'd include the project (thinking the `anvil-csharp-core` part) in your solution and be able to easily toggle flags and recompile through there.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: yea, all good

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: yes...we...
I joke, it is the consistent style of the library.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
jkeon: Ah ok, I had originally just read that it was checking if it was one rather than the start of one. I feel like this list will be updated pretty infrequently so it should be fine to just keep as the hashset and add to it as needed.

mbaker3: Technically there's a risk that someone else defines types with the same name in the same namespace but a different assembly and we should check the source assembly of the type. That's just stupid behaviour though and I don't think we need to guard against it. This solution looks good.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -179,10 +202,19 @@ public void LogFormat(LogType logType, UnityEngine.Object context, string format
                 return;
             }
 
-            SendToLogger(context, LogTypeToLogLevel(logType), string.Format(format, args));
+            // Make UnityEngine.Debug.Assert failures throw an exception instead of just logging an error
+            StackTrace stackTrace = new StackTrace(fNeedFileInfo: true);
+            MethodBase assertMethod = stackTrace.GetFrame(2)?.GetMethod(); // Skip this function and UnityEngine.Logger
+            if (assertMethod != null && assertMethod.Name == "Assert" && assertMethod.DeclaringType.FullName == "UnityEngine.Debug")
```
mbaker3, line 208: A little bit safer
```suggestion
            if (assertMethod != null && assertMethod.Name == nameof(Debug.Assert) && assertMethod.DeclaringType.Assembly.ToString() == typeof(UnityEngine.Debug).Assembly.ToString())
```

Just for demonstrative purposes. You probably want to cache that `typeof`

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: Check out my proposal at https://github.com/decline-cookies/anvil-unity-core/commit/156e566a8252149c1e044358876c3fd6037e953c

Summary:
 - Publish two DLLs one for DEBUG builds and one for Release builds
    - The debug variant has a `*-debug` suffix in the file name
 - Have Unity auto select them using the [Define Constraints](https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html#conditional-assembly) feature
 - I've gone ahead and assumed that you only want `UNITY_ASSERTIONS` enabled when `DEBUG` is also enabled. The Unity player settings allow assertions to be enabled in release builds but 99% of the time that's not what you want. It's not worth creating a permutation for every possible define combination.

The downside to this approach is that you have to run two different build configs manually anytime you change the logging dll.  For now we can just manually enforce that if one DLL changes the other DLL changes as well. In the future, if this approach is accepted and after #72 we can have the DLLs generated and committed when PRs are merged/posted automatically.

If you guys like this then I can take a similar approach with the `anvil-csharp-core` logging dll

Thoughts?

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: The problem with changing this higher up in the `Logger` is that the normal resolution of the file name is resolved for free at compile time using the `[CallerFilePath]` parameter attribute. I wouldn't want to give that up as it represents the vast majority of logging calls.

What if we constructed a fake path based on the reflected type name?
Ex: If you need to use the reflected type you could pass in a "file path" of
`<Virtual Runtime>/{fullTypeMame}.generated`

maybe even replace `.` in the namespace with `/` or group the whole namespace into a single directory name. Ex: `<Virtual Runtime>/My.Name.Space/MyClass.generated`


If the type name on its own isn't enough the log handler(s) that you're using should be configurable to prepend different/more information to a log message if needed for a specific use case.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
tjvezina: I'm in the "that's just stupid behaviour" boat, but even if you did add to `UnityEngine.Assertions` or something, that sounds like more code that should be skipped in determining log context anyway.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -267,47 +299,52 @@ private void ProcessPendingLogs()
...
+
+            string callerPath = callerFrame?.GetFileName() ?? callerMethod?.ReflectedType.FullName;
 
             Logger logger = context != null ? Log.GetLogger(context) : Log.GetStaticLogger(ResolveContextFromMethod(callerMethod));
-            logger.AtLevel(logLevel, message, callerFrame.GetFileName(), callerMethod?.Name, callerFrame.GetFileLineNumber());
+            logger.AtLevel(logLevel, message, callerPath, callerMethod?.Name, callerFrame?.GetFileLineNumber() ?? 0);
         }
 
-        private (StackFrame callerFrame, MethodBase callerMethod) ResolveCaller(int skipFrames)
+        private (StackFrame callerFrame, MethodBase callerMethod) ResolveCaller(StackTrace stackTrace = null)
```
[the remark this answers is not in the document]
tjvezina: I'm fine with either solution - the thought was there may be other situations where a custom stack trace should be used, but I can't think of any so I'll pass the exception instead.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
tjvezina: Now _this_ is my JS showing 😛 Can't believe I forgot to at least use `nameof(...)`

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
tjvezina: The fake "type path" seems the most functional, but boy what a code smell. I might simplify things and leave the fallback as `Name` instead of `FullName` - in some cases it's weirdly abstract like `(Debug|...` but other times `(Anvil.Unity.Logging.UnityLogListener|...` is too much.

For what it's worth, I tried checking for null/empty `callerPath` in the handler instead and falling back on `callerDerivedTypeName` instead (i.e. the owner of the Logger instance) but that doesn't always match the log context - for example, if I create a second instance of `UnityLogListener` in some custom script, I want the log context to be `UnityLogListener` as that's what threw the exception, but Unity internally calls `ILogListener.LogException` with the `context` object set to that custom script, so the Logger instance ends up belonging to that.

Anyway, I think simply falling back on `DeclaringType.Name` is simple and easy, and there are ways to get more context if you need it.

tjvezina: Here's a thought - if the fallback fails too, should there be some default text instead of just blank? For example, if the entire call stack happens to use `[Exclude]` attributes, instead of the context looking like `(|:0)` it could be `(<unknown>|<unknown>:0)` or similar?

Also, is it more readable for the line number to be grouped with the file name instead?
```
(MyScript:42|OnEnable) Enabling my object!
```

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
tjvezina: I'm on board with this solution - the assumptions are, the logging DLLs are a necessary evil to satisfy Unity, so we'll treat them as regular parts of Anvil, and some day they'll hopefully be built for us.

I'm not sure CI is even the fix for that? It's more like, when I make a change to a logging system file, I wish it would auto-compile the DLL right there in Unity. But at least external builds would ensure they stay up to date in the repo.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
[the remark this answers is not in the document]
tjvezina: This led me to catch something I left in by accident anyway - I was pulling the full StackTrace in `LogFormat` to check for `Debug.Assert` errors, when it really only needs to say `new StackFrame(2)`, so the last commit improves logging performance too. :)

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
tjvezina: I ran into a problem with Define Constraints - Unity is telling me that `!DEBUG` is invalid. You can definitely use `!` to invert conditions, and use any built-in or custom define symbol... except `!DEBUG`, then it gives an invalid condition error and removes it.

Beyond that, there were issues with adding a second DLL with a different assembly name (anvil-unity-logging-debug) which throws off the helpers looking for an assembly called "anvil-unity-logging". I tried instead keeping the assembly name the same in both and simply renaming one with "-debug" after building, but then Unity refuses to load a DLL who's name doesn't match its assembly 😠

Maybe we just leave the DLLs as release builds in the repo, so no one accidentally makes a release build with debug libraries, and during development you just have to make local debug builds yourself?

jkeon: I will defer to your and @mbaker3 's judgement on this as I'm not super well versed. 

If `!DEBUG` doesn't work, are you saying that something like `!ANVIL_DEBUG` would work? 

Because then maybe we could do something like this in Anvil core code? 

```csharp
#ifdef DEBUG
#define ANVIL_DEBUG
#endif
```

Although I think what I just wrote is C/C++ specific and you'd have to make it an Editor script that modified the player settings. 

Anyway just food for thought. 

I think ultimately we just don't want to have to do anything annoying or manual just regular development because it won't get done. If it's an annoying/manual one time thing when making changes to the logging projects/dll I think that's fine since it's pretty infrequent.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
jkeon: Fallback:

Yeah it would be nice to see that the fallback failed in the log. When I see `(|:0)` it makes me think that there was some weirdness with the execution and what i was trying to log was hidden in some DLL or I didn't have permissions or whatever. It looks like a glitch rather than something I can investigate and fix. 

`(<file-failed>:0|<function-failed>)` would probably prompt me to start digging. 

Readability:

Yes! I like that. `(FILE:LINE|FUNCTION) MESSAGE` makes sense to me.

mbaker3: Unqualified name:
Yea, don't use fully qualified name. Most of the time you can quickly figure out any type name ambiguity based on the context of the log message or a quick check of the types with the same name. Prefer easier to scan logs over saving the odd bit of time wasted checking multiple types with the same name.

Also, in Unity you can just check the rest of the stack attached to the message and get exactly what you want.

Fallback:
Agreed that having some placeholder text is a good idea. I think I prefer `<unknown>` but I'm not picky on this.

Readability:
Disagree on the default changing...not just because I came up with it 😆 
Right now it reads widest scope to narrowest scope. If people prefer something different on their projects they can change the formatting on the handlers for their specific application.

I vote it stays as is but it's not just up to me!

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: > I'm not sure CI is even the fix for that? It's more like, when I make a change to a logging system file, I wish it would auto-compile the DLL right there in Unity. But at least external builds would ensure they stay up to date in the repo.

There isn't really a way to keep the DLLs up to date. It could always just be a check to make sure the checksum of the DLLs match what you would get if you compiled the source proposed in the merge. That would catch the same issue. A problem for another day.

> I ran into a problem with Define Constraints - Unity is telling me that !DEBUG is invalid. You can definitely use ! to invert conditions, and use any built-in or custom define symbol... except !DEBUG, then it gives an invalid condition error and removes it.

I didn't have this issue. I did notice you had to re-select the DLL to get the inspector to refresh after applying the constraint. I also saw a red ❗ in the inspector but that seemed to indicate that the condition was not met vs a ✔️ that was shown when the condition was met on the other DLL.
For context I was using Unity 2021.3.5. Maybe there are issues with other versions?

> Beyond that, there were issues with adding a second DLL with a different assembly name (anvil-unity-logging-debug) which throws off the helpers looking for an assembly called "anvil-unity-logging". I tried instead keeping the assembly name the same in both and simply renaming one with "-debug" after building, but then Unity refuses to load a DLL who's name doesn't match its assembly 😠

Not sure I understand this one. Could the helpers not just conditionally look for "-debug" in the name when `#if DEBUG`?

tjvezina: > I also saw a red ❗ in the inspector but that seemed to indicate that the condition was not met vs a ✔️ that was shown when the condition was met on the other DLL.

If you hover over the ❗ it says invalid condition - if you type anything else like `!UNITY_ASSERTIONS` or `!MY_SYMBOL` or `!DEBU` it shows the ✔️ but the second you type `!DEBUG` it changes to the ❗ and when you hit [Apply], it just removes it from the list.

> Could the helpers not just conditionally look for "-debug" in the name when #if DEBUG?

Sure they could, but I didn't want to keep adding layers of future "what was this for?" weirdness - I can keep trying to get it working if you prefer having both debug and release DLLs, but at this point I would vote to keep the implementation simpler and just have release DLLs in the repo.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
jkeon: Readability: I don't feel too strongly on this to cast a serious vote. Widest to narrowest scope is a valid argument.

tjvezina: My only argument for changing the log context order is that I didn't realize until this PR that the first part was the file name - I assumed it was the Logger's owner, or at least the type that declares the method. I suggested `(File:Line|Method)` because `File:Line` is a familiar format, and makes it more obvious that it's saying "this file at this line". (If I could figure out how, I would skip the line number when the file path falls back on `DeclaringType.Name` but that's a whole other issue).

If instead the goal is to sort by scope, I would trade the bar for a dot: `(File.Method:Line)` so they don't look like separate pieces, otherwise I read it as ("file/type name" | "method at line #")

If you're still not convinced, I'm fine leaving it as-is, just wanted to ask the question and share my rationale!

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: I'm not seeing `!DEBUG` remove on apply. See screenshot. The ❗ does seem to mean the condition isn't met. The reason you're seeing the ❗ with `!DEBUG` is because that condition isn't met. Trying putting in another condition that is clearly false in the current context. You'll get the same visual treatment.
image
image

> Sure they could, but I didn't want to keep adding layers of future "what was this for?" weirdness

I understand now. It's `LogProjectUtils` that's having trouble auto updating references in the csproj automatically.

> ...I would vote to keep the implementation simpler and just have release DLLs in the repo.

yea, this has drifted out of scope. I think it's worth investigating some other day. Will create a reference issue.

tjvezina: You're right - when you hover the ❗ the message is "Define constraint is incompatible" (and the one at the top of the list is "One or more define constraints is invalid or incompatible") which I took to mean "this is wrong, fix it" but apparently means the condition is just false right now 😆

I can't get it to remove `!DEBUG` again, maybe it reimported the DLL once or something before, but yeah it does appear to be working as expected. Happy to leave this to another task/day though.

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
mbaker3: Your rationale is valid. Feel free to change the handlers to a different default format. Unless you've already done the work let's submit the change in a separate PR though. (#79)

> (If I could figure out how, I would skip the line number when the file path falls back on DeclaringType.Name but that's a whole other issue).

Not really possible today but you could configure to always include both in your application if you wanted. Probably not ideal.

> If instead the goal is to sort by scope, I would trade the bar for a dot: (File.Method:Line) so they don't look like separate pieces, otherwise I read it as ("file/type name" | "method at line #")

The trouble with a `(File.Method:Line)` is that it's misleading since the file name is not necessarily close to the type name. I lived with this for a bit in a project and it wasn't nice.

PR #81, opened 2022-10-06
Hotfix id static reset

mbaker3:
Leverage the changes in https://github.com/decline-cookies/anvil-csharp-core/pull/114 to properly reset `ID`'s static state.

#### Tag Alongs
 - Rename `StaticStateResetHelper` to `Logging_StaticStateResetHelper` to match the naming convention established in this PR and reduce confusion when trying to find the right static state reset helper.

Instances that listen to `ID.OnIDLimitGlobalWarning` that do not or are unable to remove their handler between executions will be incorrectly notified in the editor when domain reload is enabled.

Using `[RuntimeInitializeOnLoad]` the static state of `ID` is reset between run sessions.
A log message is also emitted to help debugging just in case there is a race condition where a static type or method executed with `[RuntimeInitializeOnLoad]` is trying to listen to the event.

None

 - https://github.com/decline-cookies/anvil-csharp-core/pull/114

 - [ ] Yes 
 - [x] No


jkeon (approved):

mbaker3:
Recompiled dll due to conflict on `main`

mbaker3:
Just added a quick fix to a compile error that was introduced in #82

tjvezina (approved):

PR #82, opened 2022-10-06
Logging polish (part 2)

tjvezina:
Omit line number from log context when zero

Logs from managed code have line number zero in the context, ex: `(UnityLogListener:0|.ctor) message`

Logs from managed code now omit the line number, ex: `(UnityLogListener|.ctor) message`

 - None

 - [ ] Yes
 - [X] No

### Related
- https://github.com/decline-cookies/anvil-csharp-core/pull/115

### DIFF
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogHandler.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogHandler.cs
@@ -26,9 +26,14 @@ namespace Anvil.Unity.Logging
         {
             string callerFile = Path.GetFileNameWithoutExtension(callerPath);
 
-            string context = (callerLine > 0 ? $"{callerFile}:{callerLine}|{callerName}" : $"{callerFile}|{callerName}");
-
-            message = $"({context}) {message}";
+            if(callerLine > 0)
+            {
+               message = $"({callerFile}:{callerLine}|{callerName}) {message}"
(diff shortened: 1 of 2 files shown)


jkeon (approved):

Scripts/Runtime/Logging/.CSProject/UnityLogHandler.cs
```diff
@@ -24,7 +24,11 @@ public class UnityLogHandler : ILogHandler
             string callerName,
             int callerLine)
         {
-            message = $"({Path.GetFileNameWithoutExtension(callerPath)}:{callerLine}|{callerName}) {message}";
+            string callerFile = Path.GetFileNameWithoutExtension(callerPath);
+
+            string context = (callerLine > 0 ? $"{callerFile}:{callerLine}|{callerName}" : $"{callerFile}|{callerName}");
+
+            message = $"({context}) {message}";
```
mbaker3, line 31: let's add the message in one step to avoid the extra string allocation.

```suggestion
            if(callerLine > 0)
            {
               message = $"({callerFile}:{callerLine}|{callerName}) {message}"
            }
            else
            {
               message = $"({callerFile}|{callerName}) {message}"
            }
```

mbaker3 (changes requested):

PR #83, opened 2022-10-07
`anvil-unity-logging.csproj` update from .NET Framework to Standard/Core

tjvezina:
Updated the logging project to a modern `.csproj` targeting Standard 2.1 / Core 3.1.

- In line with anvil-csharp-core project requirements
- Cleaner SDK-style .csproj, no longer need Properties/AssemblyInfo.cs or namespace prefix in `LogProjectUtils`
- Adds new *.deps.json required by modern .NET DLLs

- Cleaner project files, easier to work with

 - https://github.com/decline-cookies/anvil-csharp-core/pull/116

 - [ ] Yes
 - [X] No

### Notes
To ensure these changes aren't just working for me, I would suggest checking out both the C# and Unity branches for these PR's and confirming that the new logging DLLs still function the same.


Scripts/Runtime/Logging/.CSProject/Logging_StaticStateResetHelper.cs.meta
```diff
@@ -1,3 +0,0 @@
-fileFormatVersion: 2
-guid: 87c32da738604588800033dca9be761f
-timeCreated: 1657901177
```
tjvezina, line 3: I can restore it if necessary but - did this need a .meta file? It's within the .CSProject so Unity doesn't see it right?

jkeon (approved):

Scripts/Runtime/Logging/.CSProject/Logging_StaticStateResetHelper.cs.meta
mbaker3: You're right. Should probably ignore all meta files under the .CSProject directory to avoid this accident in the future. (new PR)

tjvezina: I mean, normally that's not a concern - Unity can't see this folder to generate the .meta files in the first place. Maybe this file started in Unity and was moved up? Either way, harmless.

mbaker3: Yea or Rider auto created it.

PR #84, opened 2022-10-07
NUnit test framework setup

tjvezina:
Setup for Unity's test framework, to allow writing and running NUnit tests.

- A lack of unit testing to guard against regression

 - None

 - [ ] Yes 
 - [X] No

### Related
- https://github.com/decline-cookies/anvil-csharp-core/pull/117

---

By the simple addition of this .asmdef file, Unity will pick up any tests found until this editor-only assembly and display them in the "Test Runner" window (Window > General > Test Runner).

image2

A similar Tests folder + asmdef could be set up for anvil-unity-dots, but I leave that up to others if deemed valuable. Unity also allows play mode tests, but until such time as a need for those arises, only edit mode tests were set up here.

### NUnit
[NUnit Reference](https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertions.html)
[Unity Test Framework](https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/index.html)

Writing tests is as easy as adding a `[Test]` attribute to a function, and using NUnit's `Assert.That(...)` system to test functionality. 

Here's an example script:
```cs
using NUnit.Framework;
using Anvil.CSharp.Logging;

namespace Anvil.Unity.Tests
{
    public static class LoggerTests
    {
        [Test]
        public static void TestStaticLogger()
        {
            Logger logger = Log.GetStaticLogger(typeof(LoggerTests));

            Assert.That(logger.DerivedTypeName, Is.EqualTo(nameof(LoggerTests)));
        }
    }
}
```

### DIFF
--- a/Scripts/Editor/Tests/anvil-unity-tests.asmdef
+++ b/Scripts/Editor/Tests/anvil-unity-tests.asmdef
@@ -1,0 +1,16 @@
+{
+    "name": "anvil-unity-tests",
+    "rootNamespace": "Anvil.Unity.Tests",
+    "references": [],
+    "includePlatforms": [
+        "Editor"
+    ],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": [],
+    "noEngineReferences": false
+}
--- a/Scripts/Editor/Tests.meta
+++ b/Scripts/Editor/Tests.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 38909e6e0202edf47a53bba05b459fb6
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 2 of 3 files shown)


Scripts/Editor/Tests/anvil-unity-tests.asmdef
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "anvil-unity-tests",
+    "rootNamespace": "Anvil.Unity.Tests",
+    "references": [],
```
jkeon, line 4: We probably want to reference the Anvil asmdefs. 

`anvil-unity-core-editor`
`anvil-unity-core-runtime`
`anvil-csharp-core`

jkeon (approved):
Looks good, one minor change

Scripts/Editor/Tests/anvil-unity-tests.asmdef
tjvezina: By default, assembly definitions automatically reference other assemblies as needed, so it doesn't require an explicit list (i.e. I don't have to tell it to use the nunit DLL either). I figured that was fine for a testing assembly that isn't referenced by anything else.

jkeon: Oh that's good to know, fair enough!

mbaker3 (approved):
Approved assuming the screenshot was just an example and you purposely didn't include any actual tests in the PR.

tjvezina:
Exactly, no actual unit tests here yet. I suppose the folder structure within the `Tests/` folder is open for discussion as well, but for now it's just empty.

PR #86, opened 2022-10-11
Ignore .meta files in logging DLL project directory

mbaker3:
Ignore `.meta` files in the logging DLL project director.
This directory doesn't get imported by Unity so it shouldn't have any `.meta` files.

`.meta` files may be accidentally added to the `.CSProject` logging directory.

`.meta` files are ignored in `.CSPRoject` all all sub-directories.

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ $RECYCLE.BIN/
 
 # Mac desktop service store files
 .DS_Store
+
+# Logging project (not imported by Unity)
+/Scripts/Runtime/Logging/.CSProject/**/*.meta


tjvezina:
While you're at it, I noticed anvil-csharp-core ignores `anvil-csharp-logging.deps.json`, which now exists for the Unity logging DLL too - the internet strongly recommended keeping that file, but I'm not sure it's necessary for Anvil, should it be removed and ignored here as well, or should the anvil-csharp-logging.deps.json be un-ignored?

jkeon (approved):
Looks good, but I'll leave you to answer @tjvezina's comment

mbaker3:
> While you're at it, I noticed anvil-csharp-core ignores `anvil-csharp-logging.deps.json`, which now exists for the Unity logging DLL too - the internet strongly recommended keeping that file, but I'm not sure it's necessary for Anvil, should it be removed and ignored here as well, or should the anvil-csharp-logging.deps.json be un-ignored?

Done: https://github.com/decline-cookies/anvil-csharp-core/pull/121

PR #87, opened 2022-10-13
Implement PreventAutoRefreshWhilePlaying

mbaker3:
Add option to prevent Unity from refreshing the asset database (import/compile) while the editor is playing.
Option located at `Anvil -> Prevent Asset Refresh While Playing`

Recent versions of Unity have either broken or removed the preference that controlled this (`Preferences -> General -> Script Changes While Playing`)

Changing code while the editor is running, even when `Recompile After Finished Playing` triggers an import operation that blocks input.

If enabled, prevents unity from refreshing the asset database (import/compile) while the editor playing.

None

 - None

 - [ ] Yes 
 - [x] No - The option is disabled by default

### DIFF
--- a/Scripts/Editor/AssetImport/PreventAutoRefreshWhilePlaying.cs
+++ b/Scripts/Editor/AssetImport/PreventAutoRefreshWhilePlaying.cs
@@ -1,0 +1,108 @@
+using Anvil.CSharp.Logging;
+using UnityEditor;
+
+namespace Anvil.Unity.Editor.AssetImport
+{
+    /// <summary>
+    /// Prevents the asset database from refreshing while the editor is in play mode.
+    /// This means that any changed assets aren't imported and changed source isn't recompiled.
+    ///
+    /// NOTE: When this is enabled it overrides the behaviour of
+    /// Preferences -> General -> Script Changes While Playing
+    /// (In versions of Unity that include that option)
+    /// </summary>
+    [InitializeOnLoad]
+    public static class PreventAutoRefreshWhilePlaying
+    {
+        private const string MENU_PATH = "Anvil/Prevent Asset Refresh While Playing";
+        private const string PREFSKEY_IS_ENABLED = MENU_PATH + ":IsEnabled";
+
+        private static readonly Logger s_Logger = Log.GetStaticLogger(typeof(PreventAutoRefreshWhilePlaying));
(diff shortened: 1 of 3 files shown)


Scripts/Editor/AssetImport/PreventAutoRefreshWhilePlaying.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        [MenuItem(MENU_PATH, true)]
+        private static bool ToggleEnabled_Validator()
+        {
+            // Make sure the menu is showing the correct checked state
+            Menu.SetChecked(MENU_PATH, IsEnabled);
+            return true;
+        }
+
+        private static void EditorApplication_PlayModeStateChanged(PlayModeStateChange state)
```
jkeon, line 67: Does this get called multiple times with the domain reloading turned off? Not sure if it affects Editor static events the same way or not but we should check and guard against it if it does.

jkeon (approved):
Looks good, one thing to double check before merging

Scripts/Editor/AssetImport/PreventAutoRefreshWhilePlaying.cs
mbaker3: Not sure the concern you're worried about.

If the concern is that this gets called multiple times for a `state` during a single transition from play to edit (or vice versa) then it's fine. The `Allow/PreventAutoRefresh()` methods are well protected against being called multiple times. They have to be because `AssetDatabase.DisallowAutoRefresh();` and `AssetDatabase.AllowAutoRefresh();` are a bit dangerous. They increment and decrement a private counter so if you get these calls out of balance you can really break the state of the editor.

If the concern is that this doesn't called when domain reload is turned off, it does. The logs emitted by `Allow/PreventAutoRefresh()` are shown on every transition.

jkeon: My concern was that the static constructor would get called multiple times during development after every recompile but there wasn't ever a removal from that listener. 

Spend a few days without closing Unity and you have a build up of the function getting hammered by multiple listeners. 

I know you've gated the heavy work which is great, I was just concerned about a hidden runaway slow down. 

I did a quick test and it doesn't seem to be the case. Logs in the function only trigger once so we're good. 

You're free to merge.

PR #89, opened 2022-10-13
Misc minor fixes

tjvezina:
- Added missing .asmdef references to anvil-unity-tests.asmdef
- Added missing self-cloding slash in `<see />` tags
- Changed anvil-unity-logging.csproj to target .NET Standard 2.1 only
- Added `RectUtilTests` to test Unity unit tests :)

- None

 - None

 - [ ] Yes
 - [X] No


Scripts/Editor/Tests/anvil-unity-tests.asmdef
```diff
@@ -1,7 +1,11 @@
 {
     "name": "anvil-unity-tests",
     "rootNamespace": "Anvil.Unity.Tests",
-    "references": [],
+    "references": [
+        "anvil-unity-core-runtime",
+        "anvil-unity-core-editor",
+        "anvil-csharp-core"
+    ],
```
mbaker3, line 8: Glad this works today. Might need to split into Editor and Runtime tests some day in the future.
Seems to be common in other packages
image

mbaker3 (approved):
Looks good!
One comment about potentially splitting up the tests some day.

Thanks for fixing those open `<see />` tags!

Scripts/Editor/Tests/anvil-unity-tests.asmdef
tjvezina: Unity does support "Play mode" unit tests as well, which I assume is the distinction here - I only bothered setting up edit mode tests for now, but yeah it should be possible to do the same setup in the runtime folder, and do some fancy in-game testing too.

jkeon (dismissed):

Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
```diff
@@ -307,10 +306,15 @@ private void SendToLogger(UnityEngine.Object context, LogLevel logLevel, string
...
+            string callerFilePath = callerFrame?.GetFileName();
+            int callerLineNumber = callerFrame?.GetFileLineNumber() ?? 0;
 
-            Logger logger = context != null ? Log.GetLogger(context) : Log.GetStaticLogger(ResolveContextFromMethod(callerMethod));
-            logger.AtLevel(logLevel, message, callerPath, callerMethod?.Name, callerFrame?.GetFileLineNumber() ?? 0);
+            Logger logger = context != null
+                ? Log.GetLogger(context)
+                : Log.GetStaticLogger(callerMethod?.ReflectedType ?? typeof(UnknownContext));
+
+            logger.AtLevel(logLevel, message, callerMethodName, callerFilePath, callerLineNumber);
```
tjvezina, line 317: This change came from the rework in https://github.com/decline-cookies/anvil-csharp-core/pull/123, namely the inclusion of the logger's `ownerTypeName` in the log context.

Previously, if no file path was available (i.e. for DLL code) it would fall back on the method's owner type. This no longer makes sense as the caller type is now also included in the log context, and furthermore Unity's `context` object passed to `LogException` would not actually be the source of the exception, but the last non-DLL function in the stack. For example, creating a second `UnityLogListener` instance from a "Demo.cs" script would result in:
```
(Demo..ctor|UnityLogListener) System.Exception: There can only be one instance [...]
```
...but with these changes, the missing file path is left blank, and the context falls back on the calling method's owner type, resulting in:
```
(UnityLogListener..ctor|<unknown>) System.Exception: There can only be one instance [...]
```

jkeon: 👍

jkeon (approved):

Issue #90, opened 2022-10-20
UnityLogListener - Remove Unity Thread Marshalling


mbaker3:
**Depends On:** https://github.com/decline-cookies/anvil-csharp-core/issues/126

Once the dependent work is complete to make `Log`/`Logger` calls thread safe remove the unity main thread marshalling from `UnityLogListener`

PR #91, opened 2022-10-22
UnityLogListener - Make thread safe + bug fixes

mbaker3:
All logging calls into `UnityLogListener` are now thread safe. This is an interim fix until the proper work of making `Log` thread safe is done https://github.com/decline-cookies/anvil-csharp-core/issues/126

Also fixes an issue where the main thread pump game object wasn't getting run in the Unity editor when playing.

#### Tag Alongs
 - Add comment to reference changes needed when Log is made thread safe. #90 
 - `SKIPPED_STACK_FRAME_TYPES` now supports methods and type detection. This allows finer grained control where we want to skip specific methods but not the whole type in a stack

If a developer calls `Debug.Log` (and friends) from within a job with the Burst compiler disabled the `Log.IsHandlingLog` check will throw an error because two threads are trying to log at once. This is because while `Debug.Log` is thread safe, `UnityLogListener.LogFormat/LogException` are not.

When the burst compiler is enabled this error is not encountered because logs emitted from burst compiled code bypass `LogFormat/LogException` and go straight through to the `Application.logMessageReceivedThreaded` event.

#### Main thread Pump in Editor
When transitioning from editor to play mode the main thread pump game object is removed from the player loop and left in editor space. This meant that logs received through `Application.logMessageReceivedThreaded` were piling up in the `pendingLogs` queue and never getting sent through the `Log`.

 - `Debug.LogException/LogFormat` off main thread calls are now marshalled through the pending logs queue just like the burst logs.
 - Refactor the log context building logic to make it easier to package up and either send to Log immediately or later. This also allows log context to be built on thread rather than only on the main thread.

#### Main thread Pump in Editor
Non-main thread logs now correctly pass through to handlers. The pump is re-created during the `SubsystemRegistration` phase at play time.

#### Tag Along
 - `SKIPPED_STACK_FRAME_TYPES` now supports methods and type detection. As a result, the call site at `UnityEngine.Application.CallLogCallback` is now skipped

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
+++ b/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs
@@ -7,11 +7,13 @@ using StackTrace = System.Diagnostics.StackTrace;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections;
+using System.Threading;
 using UnityEngine.Scripting;
 using Logger = Anvil.CSharp.Logging.Logger;
 
 namespace Anvil.Unity.Logging
 {
+    // TODO: #90 - Remove unity main thread marshalling once Log/Logger are made thread safe
     /// <summary>
     /// An implementation of <see cref="ILogListener"/> for the Unity <see cref="Debug"/> logger.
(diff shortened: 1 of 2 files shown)


jkeon (approved):
🎉

PR #93, opened 2022-12-05
RectUtil - Add CreateFromPoints 4 point overload

mbaker3:
Add an overload to `RectUtil.CreateFromPoints` that accepts 4 points.

A developer may only provide 2 arbitrary points when creating a rectangle with the utility.

A developer may provide 2 or 4 arbitrary points when creating a rectangle with the utility.

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Core/Util/RectUtil.cs
+++ b/Scripts/Runtime/Core/Util/RectUtil.cs
@@ -1,5 +1,6 @@
 
 
+using Unity.Mathematics;
 using UnityEngine;
 
 namespace Anvil.Unity.Core
@@ -10,7 +11,7 @@ namespace Anvil.Unity.Core
     public static class RectUtil
     {
         /// <summary>
-        /// Creates a <see cref="Rect"/> from two arbetrary points
+        /// Creates a <see cref="Rect"/> from two arbitrary points
         /// </summary>
         /// <param name="point1"></param>
         /// <param name="point2"></param>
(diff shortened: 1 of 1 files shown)


Scripts/Runtime/Core/Util/RectUtil.cs
```diff
@@ -46,5 +47,28 @@ public static Rect CreateFromPoints(Vector2 point1, Vector2 point2)
...
+
+            float minX = math.cmin(x);
+            float minY = math.cmin(y);
+
+            float maxX = math.cmax(x);
+            float maxY = math.cmax(y);
+
+            return Rect.MinMaxRect(minX, minY, maxX, maxY);
+        }
+
```
tjvezina, line 72: 🚲 
```suggestion
```

tjvezina (dismissed):
> ### What is the new behaviour?
> A developer may only provide 4 arbitrary points when creating a rectangle with the utility.

I assume this was meant to say "may provide 2 or 4".

It sounds like the points given are not necessarily the corners, but any arbitrary points - would it be more accurate then to call this `GetBoundingRect` instead of `CreateFromPoints`? Would a more general function with `params Vector2[] points` be useful?

tjvezina:
These changes seem to be included in #97 too

mbaker3:
> I assume this was meant to say "may provide 2 or 4".

yea, fixed.

> It sounds like the points given are not necessarily the corners, but any arbitrary points - would it be more accurate then to call this GetBoundingRect instead of CreateFromPoints? 

That's a good point. From the case I was using this for I had assumed the four points would be corners just out of order. In my haste I didn't even consider they may not represent extremes!

I'll put some thought into a name and revise this PR.

> Would a more general function with params Vector2[] points be useful?

A `params` overload is a good idea but it's worth keeping the 4 point version to avoid the heap allocation of the array.

mbaker3:
- Chose the name `CreateBoundingRect`
- Created an overload that accepts a collection of points and added a TODO to optimize further (#99)

jkeon (approved):

PR #94, opened 2022-12-05
Add `TransformExtension.SetLossyScale`

mbaker3:
Add `TransformExtension.SetLossyScale`.

Developer must manually calculate the `localScale` value for a `Transform` to achieve a given world scale(`lossyScale`).

`TransformExtension.SetLossyScale` provides a convenience method that allows developers to set the world scale value of a `Transform`

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Editor/Tests/Core/Util/TransformExtensionTests.cs
+++ b/Scripts/Editor/Tests/Core/Util/TransformExtensionTests.cs
@@ -1,0 +1,41 @@
+using Anvil.Unity.Core;
+using NUnit.Framework;
+using UnityEngine;
+
+namespace Anvil.Unity.Tests
+{
+    public class TransformExtensionTests
+    {
+        [Test]
+        public static void SetLossyScaleTest()
+        {
+            Assert.That(nameof(SetLossyScaleTest), Does.StartWith(nameof(TransformExtension.SetLossyScale)));
+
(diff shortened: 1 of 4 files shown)


Scripts/Editor/Tests/Core/Util/TransformExtensionTests.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public class TransformExtensionTests
+    {
+        [Test]
+        public static void SetLossyScaleTest()
+        {
+            Assert.That(nameof(SetLossyScaleTest), Does.StartWith(nameof(TransformExtension.SetLossyScale)));
+
+            Vector3 scale_two = new Vector3(2f, 2f, 2f);
+            Vector3 scale_two_one_two = new Vector3(2f, 1f, 2f);
```
tjvezina, line 15: What's with the underscore naming style?

mbaker3: Finding it easier to read. Particularly with very large test files. The asserts are very information dense and this breaks things up a bit.

If you and @jkeon really don't like it I can rework them to match the rest of the lib.

FWIW this is more or less the naming convention for constants and that's effectively what these are. Method level constants.

jkeon (approved):
Looks good.

PR #96, opened 2022-12-06
Add `MathExtension`

mbaker3:
Add a series of extension methods that extend the capabilities of types under the `Unity.Mathematics` library.

There are some common operations with `Unity.Mathematics` types that are cumbersome or error prone to type.

The extension methods defined under `MathExtension` are as follows:
 - `GetInverse()`/`GetInverseSafe()` - Calculate the inverse of a `float3`
 - `IsApproximately`/`IsApproximatelySafe` - Calculate equality between two floating point values using the standard Anvil threshold defined by `MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE`
    - The safe variant resolves NaN using the same logic as `IsEqualOrNaN` and also handles equality of infinite values.
 - `ToSignedInfinite` - Returns signed infinite values for each component of a floating point value
 - `IsEqualOrNaN` - Performs a standard equality check but also resolves `float.NaN == float.NaN` to `true` (unlike standard equality checks).

#### Safe vs Fast Suffix Conventions
The extension methods in this PR all use the "Safe" suffix but this isn't an arbitrary choice. The guidelines I've been using are as follows:
 - **Safe** - Adds additional guards against edge case invalid values (NaN, Infinity, etc..) at the cost of performance. The developer can choose to use this method if they haven't already guarded against anomalous values.
 - **No Suffix** - Assumes input values are valid but doesn't make assumptions about the nature of the values beyond that.
 - **Fast** - Makes assumptions about the nature of the values provided to perform the operation more quickly.

**Example:**
The following is an example using invert. This is not an actual implementation in this PR but does help show the relationship between the suffix choices. This is roughly based on the difference between Unity's `math.inverse` and `math.fastinverse` methods.

 - `InvertMatrixSafe` - Accepts matrices that contain NaN and Infinite values, substituting in valid values (usually 0) and inverting the matrix using a general purpose algorithm. The developer can confidently throw any matrix at this and get a valid output at the cost of performance.
 - `InvertMatrix` - Will not guard against NaN and infinite meaning they may persist into the inverted matrix or propagate to other values in the matrix. The developer needs to ensure input values are valid (or check the output) but benefits from higher performance than the Safe variant.
 - `InvertMatrixFast` - In addition the the assumptions made in `InvertMatrix` the matrix is assumed to have an [orthonormal basis](https://en.wikipedia.org/wiki/Orthonormal_basis) that allows an optimized/shortcut algorithm perform the operation. The developer needs to be sure that the input matrix has the properties assumed by the method and benefits from the highest performance of the three variants.

None

 - #95 
 - https://github.com/decline-cookies/anvil-csharp-core/pull/136

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Core/Util/MathExtension.cs
+++ b/Scripts/Runtime/Core/Util/MathExtension.cs
@@ -26,7 +26,7 @@ namespace Anvil.Unity.Core
         }
 
         /// <summary>
-        /// Get the inverse of a <see cref="float3"/> with any <see cref="float.NaN"/> or infinite components set to 0
+        /// Get the inverse of a <see cref="float3"/> with any <see cref="float.NaN"/> or infinite components set to 0.
         /// </summary>
         /// <param name="value">The value to get the inverse of.</param>
         /// <returns>The inverted value</returns>
@@ -48,7 +48,7 @@ namespace Anvil.Unity.Core
         /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
         /// </returns>
         /// <remarks>
-        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximately"/>.
+        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximatelySafe"/>.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsApproximately(this float a, float b)
@@ -67,7 +67,7 @@ namespace Anvil.Unity.Core
         /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
         /// </returns>
         /// <remarks>
-        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximately"/>.
+        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximatelySafe"/>.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool2 IsApproximately(this float2 a, float2 b)
@@ -86,7 +86,7 @@ namespace Anvil.Unity.Core
         /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
         /// </returns>
         /// <remarks>
-        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximately"/>.
+        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximatelySafe"/>.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool3 IsApproximately(this float3 a, float3 b)
@@ -105,7 +105,7 @@ namespace Anvil.Unity.Core
         /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
         /// </returns>
         /// <remarks>
-        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximately"/>.
+        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximatelySafe"/>.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool4 IsApproximately(this float4 a, float4 b)
@@ -124,7 +124,7 @@ namespace Anvil.Unity.Core
         /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
         /// </returns>
         /// <remarks>
-        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximately"/>.
+        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximatelySafe"/>.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool3x3 IsApproximately(this float3x3 a, float3x3 b)
@@ -147,7 +147,7 @@ namespace Anvil.Unity.Core
         /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
         /// </returns>
         /// <remarks>
-        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximately"/>.
+        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximatelySafe"/>.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool4x4 IsApproximately(this float4x4 a, float4x4 b)
--- a/Scripts/Runtime/Core/Util/MathExtension.cs
+++ b/Scripts/Runtime/Core/Util/MathExtension.cs
@@ -179,7 +179,7 @@ namespace Anvil.Unity.Core
         {
             return (math.abs(a - b) < MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE)
                    | (math.isnan(a) & math.isnan(b))
-                   | (math.isinf(a) & math.isinf(b));
+                   | (math.isinf(a) & math.isinf(b) & a == b);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Core/Util/MathExtension.cs
```diff
@@ -0,0 +1,406 @@
...
+        /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
+        /// </summary>
+        /// <param name="a">The <see cref="float"/> to compare.</param>
+        /// <param name="b">The <see cref="float"/> to compare.</param>
+        /// <returns>
+        /// True if the delta between <see cref="a"/> and <see cref="b"/> is less than
+        /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
+        /// </returns>
+        /// <remarks>
+        /// Resolves infinite and NaN comparisons as false with improved performance vs <see cref="IsApproximately"/>.
```
jkeon, line 51: I think all of these should be referencing `IsApproximatelySafe`?

Scripts/Runtime/Core/Util/MathExtension.cs
```diff
@@ -0,0 +1,406 @@
...
+        /// <remarks>
+        /// Resolves infinite and NaN comparisons as true with the same logic as <see cref="IsEqualOrNaN"/> at
+        /// additional cost vs <see cref="IsApproximately"/>.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool IsApproximatelySafe(this float a, float b)
+        {
+            return (math.abs(a - b) < MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE)
+                   | (math.isnan(a) & math.isnan(b))
+                   | (math.isinf(a) & math.isinf(b) & a == b);
```
jkeon, line 182: Is the a == b check because `isinf` will return positive or negative infinity?

Scripts/Runtime/Core/Util/MathExtension.cs
```diff
@@ -0,0 +1,406 @@
...
+        /// <remarks>
+        /// Resolves infinite and NaN comparisons as true with the same logic as <see cref="IsEqualOrNaN"/> at
+        /// additional cost vs <see cref="IsApproximately"/>.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool2 IsApproximatelySafe(this float2 a, float2 b)
+        {
+            return (math.abs(a - b) < MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE)
+                   | (math.isnan(a) & math.isnan(b))
+                   | (math.isinf(a) & math.isinf(b) & a == b);
```
jkeon, line 204: Do we run into issues when only some of the elements are NaN or Infinity? 

`float2 a = new float2(5.0f, NaN)`
`float2 b = new float2(5.0f, NaN)`

We do the first check which should give... `(0.0f, NaN) < FLOATING_POINT_EQUALITY_TOLERANCE`? 
Which I think will return false because of the NaN? 

Then we do the `isnan` check but that checks both elements and 5.0f isn't NaN so they both fail. 

The above example should be true, but will return false? I wonder if instead you need to access the array indexer and run each component individually through the single `IsApproximatelySafe` call?

Scripts/Runtime/Core/Util/MathExtension.cs
```diff
@@ -0,0 +1,406 @@
...
+        /// Unlike <see cref="float2.Equals"/> this method considers <see cref="float.NaN"/> == <see cref="float.NaN"/>
+        /// to be true.
+        /// </summary>
+        /// <param name="a">The <see cref="float2"/> to compare.</param>
+        /// <param name="b">The <see cref="float2"/> to compare.</param>
+        /// <returns>True (component-wise) if the values are equal or both <see cref="float.NaN"/>.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool2 IsEqualOrNaN(this float2 a, float2 b)
+        {
+            return a == b | (math.isnan(a) & math.isnan(b));
```
jkeon, line 338: Same question above about mixed values + NaN `float2(5.0f, NaN)`

jkeon (changes requested):
Looks great, lots of questions. I kind of glazed over reading all the tests, but seems to be pretty good coverage!

Scripts/Runtime/Core/Util/MathExtension.cs
```diff
@@ -0,0 +1,406 @@
...
+            return 1f / value;
+        }
+
+        /// <summary>
+        /// Get the inverse of a <see cref="float3"/> with any <see cref="float.NaN"/> or infinite components set to 0
+        /// </summary>
+        /// <param name="value">The value to get the inverse of.</param>
+        /// <returns>The inverted value</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static float3 GetInverseSafe(this float3 value)
```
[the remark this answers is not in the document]
mbaker3: Add as needed. Had to draw the line somewhere and I haven't needed it yet.

Scripts/Runtime/Core/Util/MathExtension.cs
```diff
@@ -0,0 +1,406 @@
...
+        {
+            float3 inverse = value.GetInverse();
+            return math.select(inverse, float3.zero, math.isinf(inverse) | math.isnan(inverse));
+        }
+
+        /// <summary>
+        /// Checks if two <see cref="float"/> values are approximately equal according to the
+        /// <see cref="MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE"/>.
```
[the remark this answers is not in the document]
mbaker3: There's no concept of the Mathematics library in csharp. Doesn't see appropriate?

Scripts/Runtime/Core/Util/MathExtension.cs
mbaker3: Good catch. Fixed

Scripts/Runtime/Core/Util/MathExtension.cs
mbaker3: That's right

Scripts/Runtime/Core/Util/MathExtension.cs
mbaker3: Broke it down step by step for you below. Keep in mind that all of the operations are component-wise and we get a `bool2` out of this.

``` csharp
a-b   // 0, float.NaN
a-b < MathUtil.FLOATING_POINT_EQUALITY_TOLERANCE   // true, false
math.isnan(a) & math.isnan(b)   // false, true
(math.isinf(a) & math.isinf(b) & a == b)   // false, false

(true, false) | (false, true) | (false, false)   // true, true
```

I added a unit test for float2, float3, and tweaked float 4 to cover this. It was a good case!

Scripts/Runtime/Core/Util/MathExtension.cs
mbaker3: I think the same answer as above?

``` csharp
a == b   // (true, false)
(math.isnan(a) & math.isnan(b))   // (false, true)

(true, false) | (false, true)   // (true, true)
```

In this case, I do have a test already :)
https://github.com/decline-cookies/anvil-unity-core/blob/c0d4afce82153b6904e143c5ff6046cda53d1d7a/Scripts/Editor/Tests/Core/Util/MathExtensionTests.cs#L513

Scripts/Runtime/Core/Util/MathExtension.cs
```diff
@@ -0,0 +1,406 @@
...
+    /// A collection of extension methods for working with types under the <see cref="Unity.Mathematics"/> namespace.
+    /// (float3, int2, etc..)
+    /// </summary>
+    public static class MathExtension
+    {
+        /// <summary>
+        /// Get the inverse of a <see cref="float3"/>.
+        /// </summary>
+        /// <remarks>
+        /// Any components that are 0 will invert to <see cref="float.NaN"/>.
```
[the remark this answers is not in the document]
jkeon: TIL

Scripts/Runtime/Core/Util/MathExtension.cs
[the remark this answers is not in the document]
jkeon: Fair

Scripts/Runtime/Core/Util/MathExtension.cs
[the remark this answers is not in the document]
jkeon: Based on previous comments, fair rationale

Scripts/Runtime/Core/Util/MathExtension.cs
jkeon: Ok got it, it makes sense now.

Scripts/Runtime/Core/Util/MathExtension.cs
jkeon: Yep all good

jkeon (approved):

PR #98, opened 2022-12-06
Add QuaternionExtension

mbaker3:
Add a collection of extension methods for `quaternion`

Some common operations with `quaternion` are cumbersome.

Add extension methods to:
 - Quickly rotate 180 degrees around a fixed axis (Ex: `Rotate180AroundWorldX`)
 - `Rotate180Around` - Rotate around around a given axis
 - `ToEuler` - Convert to a euler representation

- None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Core/Util/QuaternionExtension.cs
+++ b/Scripts/Runtime/Core/Util/QuaternionExtension.cs
@@ -56,7 +56,7 @@ namespace Anvil.Unity.Core
             new quaternion(-q.value.z, q.value.w, q.value.x, -q.value.y);
 
         /// <summary>
-        /// Rotate the quaternion 180 degrees around the local z axis.
+        /// Rotate the quaternion 180 degrees around the local Z axis.
         /// </summary>
         /// <param name="q">The quaternion to rotate</param>
         /// <returns>The rotated quaternion</returns>
--- a/Scripts/Runtime/Core/Util/QuaternionExtension.cs
+++ b/Scripts/Runtime/Core/Util/QuaternionExtension.cs
@@ -1,0 +1,256 @@
+using System;
+using System.Runtime.CompilerServices;
+using Unity.Mathematics;
+using UnityEngine;
+
+namespace Anvil.Unity.Core
+{
+    /// <summary>
+    /// A collection of utilities for working with <see cref="quaternion"/>
+    /// </summary>
+    public static class QuaternionExtension
+    {
+        /// <summary>
+        /// Rotate the quaternion 180 degrees around the world X axis.
+        /// </summary>
+        /// <param name="q">The quaternion to rotate</param>
+        /// <returns>The rotated quaternion</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static quaternion Rotate180AroundWorldX(this quaternion q) =>
+            new quaternion(q.value.w, -q.value.z, q.value.y, -q.value.x);
+
+        /// <summary>
+        /// Rotate the quaternion 180 degrees around the world Y axis.
+        /// </summary>
+        /// <param name="q">The quaternion to rotate</param>
+        /// <returns>The rotated quaternion</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Core/Util/QuaternionExtension.cs
```diff
@@ -0,0 +1,256 @@
...
+using System.Runtime.CompilerServices;
+using Unity.Mathematics;
+using UnityEngine;
+
+namespace Anvil.Unity.Core
+{
+    /// <summary>
+    /// A collection of utilities for working with <see cref="quaternion"/>
+    /// </summary>
+    public static class QuaternionExtension
```
jkeon, line 11: Does this imply that we must have the `Unity.Mathematics` package anytime we use UnityCore? Is there a way to enforce that?

Scripts/Runtime/Core/Util/QuaternionExtension.cs
```diff
@@ -0,0 +1,256 @@
...
+        /// Rotate the quaternion 180 degrees around the local Y axis.
+        /// </summary>
+        /// <param name="q">The quaternion to rotate</param>
+        /// <returns>The rotated quaternion</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static quaternion Rotate180AroundSelfY(this quaternion q) =>
+            new quaternion(-q.value.z, q.value.w, q.value.x, -q.value.y);
+
+        /// <summary>
+        /// Rotate the quaternion 180 degrees around the local z axis.
```
jkeon, line 59:
```suggestion
        /// Rotate the quaternion 180 degrees around the local Z axis.
```

Scripts/Runtime/Core/Util/QuaternionExtension.cs
```diff
@@ -0,0 +1,256 @@
...
+
+        /// <summary>Returns an 180° rotated version of this quaternion around the given axis</summary>
+        /// <param name="q">The quaternion to rotate</param>
+        /// <param name="axis">The axis to rotate around (0=x, 2=z)</param>
+        /// <param name="space">The space of the axis</param>
+        /// <remarks>
+        /// This method and related methods adapted from:
+        /// https://github.com/FreyaHolmer/Mathfs/blob/446593cea11f7a25b8e73b8398d77f665039bcee/Runtime/Extensions.cs#L159-L177
+        /// </remarks>
+        public static quaternion Rotate180Around( this quaternion q, int axis, Space space ) {
```
jkeon, line 75: The `int` for the axis seems brittle. We should make/use and Enum here.

Scripts/Runtime/Core/Util/QuaternionExtension.cs
```diff
@@ -0,0 +1,256 @@
...
+        /// <param name="q">The <see cref="quaternion"/> to convert</param>
+        /// <param name="order">The rotation order</param>
+        /// <returns>The Euler values.</returns>
+        /// <remarks>
+        /// Taken from Unity.Physics.Math.cs which in turn was taken from another assembly with the comment below:
+        /// Note: taken from Unity.Animation/Core/MathExtensions.cs, which will be moved to Unity.Mathematics at some point
+        ///       after that, this should be removed and the Mathematics version should be used
+        /// https://github.com/needle-mirror/com.unity.physics/blob/master/Unity.Physics/Base/Math/Math.cs
+        /// </remarks>
+        public static float3 ToEuler(this quaternion q, math.RotationOrder order = math.RotationOrder.Default)
```
jkeon, line 96: Wild this doesn't already exist.

jkeon (changes requested):
Looks good, one suggested change and then some questions.

mbaker3:
Questions answered and request addressed

Scripts/Runtime/Core/Util/QuaternionExtension.cs
jkeon: Ah ok, if we already consider it a dependency then that's fine

jkeon (approved):

PR #106, opened 2023-01-11
QuitStateMonitor - Add

mbaker3:
Add the ability to query the state of the quitting state for an application.

Unity does not provide a way to query whether the application is currently quitting. Developers must listen to the `Application.quitting` event and maintain their own flag.

Once instantiated, `QuitStateMonitor` monitors the `Application.quitting` event and sets `QuitStateMonitor.IsQuitting` to true when the event is invoked.

Developers can now instantiate a `QuitStateMonitor` and query `IsQuitting` as needed.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (changes requested):
Not so much request changes as questions on the approach.

Why do we not automatically instantiate this statically and have it accessed that way? 

Or why not have it on `AbstractAnvilUnityApplication` since we already have the pattern of statically accessing whatever your application is as the root anyway: 

Ex: `MyApplication.ContentManager.Show(...)`

That way the bool is more likely to always reflect the true state of the application. 

Ex.
```
//Some code somewhere
Application.quitting += HandleOnQuitting1;

//Somewhere else after the previous line
QuiteStateMonitor qsm = new QuitStateMonitor();

//Somewhere else 
Application.quitting += HandleOnQuitting2;
```

In `HandleOnQuitting1`, querying the QuitStateMonitor will return false since it hasn't responded to the event yet.

mbaker3:
> Why do we not automatically instantiate this statically and have it accessed that way?

I didn't want to make this a singleton because then once the tool is used once the instance sticks around forever. The footprint is insignificant but seemed like good practice to keep things clean/simple.

For an all static implementation I didn't want this functionality to exist if you never used it. Same rationale as above regarding insignificance.

> Or why not have it on `AbstractAnvilUnityApplication` since we already have the pattern of statically accessing whatever your application is as the root anyway:
> 
> Ex: `MyApplication.ContentManager.Show(...)`

(it's a singleton right now but I get your point)
That's how you and I happen to setup applications today but it's not necessary the way it should be done. People are welcome to use dependency injection, etc..

The bulk of the code written in our game is unable to reference the application level. The use case the inspired the creation of this functionality doesn't have access to the application outside of its assembly.
What should we do if we need quit monitoring? Is it weird to have to statically access `AbstractAnvilUnityApplication`?

The concern of boolean truth is valid. What if `QuitStateMonitor` proxy's the quit event so, if you are using the monitor you are encouraged to listen through the monitor. That way, at least in your local context, the boolean will always be true. 
``` csharp
public event Action OnQuitting
{
    add => Application.quitting += value;
    remove => Application.quitting -= value;
}
```

Otherwise, there's no way I can think of to get this 100% correct but the primary use case is for an object to adapt its behaviour to the quitting state (mainly disposal). 

Maybe our best shot at it would be to subscribe to the event during subsystem registration but you still may lose out to another sub system registration method listening to `Application.Quit`. I guess I can add a warning in the docs.

You really shouldn't be polling a quit monitor in a quitting event handler. You already know you're quitting at that point but I could see a disposal getting triggered further down the stack and being unaware it was called from Application.Quit....

:trollface: I could search the stack using `Stackframe`...really though that sounds like a gross hack.

Thoughts?

jkeon:
> I didn't want to make this a singleton because then once the tool is used once the instance sticks around forever. The footprint is insignificant but seemed like good practice to keep things clean/simple.
> 
> For an all static implementation I didn't want this functionality to exist if you never used it. Same rationale as above regarding insignificance.

Sure, but the counter argument is that there is a special little class to find and learn about. I agree keeping things simple and clean is important but I think this is adding the same amount of complexity to macro understanding of the framework.

> (it's a singleton right now but I get your point)
> That's how you and I happen to setup applications today but it's not necessary the way it should be done. People are welcome to use dependency injection, etc.. 

Also fair, but I would argue that if you're using Anvil as your framework, you'd be using `AbstractAnvilUnityApplication`. 

> The bulk of the code written in our game is unable to reference the application level. 
> The use case the inspired the creation of this functionality doesn't have access to the application outside of its assembly.
> What should we do if we need quit monitoring? Is it weird to have to statically access AbstractAnvilUnityApplication?

This is a good point because of the game assembly not knowing about the App level.

> The concern of boolean truth is valid. What if QuitStateMonitor proxy's the quit event so, if you are using the monitor you are encouraged to listen through the monitor. That way, at least in your local context, the boolean will always be true.

I think this would be a good idea. I like the proposal.


> Thoughts?

I think this is a "I wouldn't do it this way" but "Your way works and has a valid use case". 
I'm cool with it but I do like the addition of proxying the event for safety.

mbaker3:
This is all good feedback and I've reworked the class to be static.
 - `QuitStateMonitor` renamed to `QuitState`
 - More likely to be the first to respond to the event
 - Provide a "safe" quitting event that guarantees the boolean has flipped

I've also updated `AbstractAnvilUnityApplication` to make use of `QuitState` and provide a convenience `IsQuitting` getter

jkeon (approved):
Nice changes, I like it!

PR #108, opened 2023-01-17
Issue #107 - Editor Menu Tool to Toggle Script Defines

### LINKED ISSUE #107: Editor menu tool to toggle compiler flags
Moved from https://github.com/decline-cookies/anvil-unity-dots/issues/106

jkeon:
Moved from https://github.com/decline-cookies/anvil-unity-dots/issues/106 and https://github.com/decline-cookies/anvil-unity-dots/pull/110

This is a simple tool to allow for quickly toggling Script Defines. 

- You have to remember what Script Defines are available
- You have to manually add/remove them from the project settings and spell them correctly
- You have to remember what other Script Defines are allowed to be on or off. 

- You can simply go to the menu item and select the script define you want to turn on or off (Status is conveyed)
- If the script define requires any other script defines to be on, they will also be turned on.
- If the script define has any other script defines that depend on it, they will also be turned off.
- The ability for downstream script defines (ex. custom safety for a piece of the project) to merge in nicely with existing framework level script defines. 
    - Ex. PATHFINDING_SAFETY requires that ANVIL_DEBUG_SAFETY is also on. ANVIL_DEBUG_SAFETY doesn't know anything about PATHFINDING_SAFETY but that's ok.  
    - When PATHFINDING_SAFETY is turned on, ANVIL_DEBUG_SAFETY will also be turned on.
    - When ANVIL_DEBUG_SAFETY is turned off, PATHFINDING_SAFETY will also be turned off.
- Added two ANVIL level script defines
    - ANVIL_DEBUG_SAFETY - General safety checks, not performance intensive
    - ANVIL_DEBUG_SAFETY_EXPENSIVE - Safety checks that are performance intensive to calculate.

- Resolves #107 

 - None

 - [x] Yes
 - [ ] No

### DIFF
--- a/Scripts/Editor/Debug/AnvilDebugSafetyScriptDefines.cs
+++ b/Scripts/Editor/Debug/AnvilDebugSafetyScriptDefines.cs
@@ -1,0 +1,68 @@
+using UnityEditor;
+
+namespace Anvil.Unity.Editor.Debug
+{
+    /// <summary>
+    /// Sets up script defines for common Anvil specific script defines
+    /// </summary>
+    public static class AnvilDebugSafetyScriptDefines
+    {
+        /// <summary>
+        /// Script Define for code safety checks.
+        /// Should not be performance intensive.
+        /// </summary>
+        public const string ANVIL_DEBUG_SAFETY = "ANVIL_DEBUG_SAFETY";
+        
+        /// <summary>
+        /// Script Define for code safety checks that are expensive to determine.
+        /// </summary>
+        public const string ANVIL_DEBUG_SAFETY_EXPENSIVE = "ANVIL_DEBUG_SAFETY_EXPENSIVE";
+        
+        
+        private const string ANVIL_DEBUG_SAFETY_MENU_PATH = ScriptDefinesToggle.MENU_PATH_BASE + "/" + ANVIL_DEBUG_SAFETY;
+        private const string ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH = ScriptDefinesToggle.MENU_PATH_BASE + "/" + ANVIL_DEBUG_SAFETY_EXPENSIVE;
+
+        //For ANVIL_DEBUG_SAFETY, we don't require anything to turn it on, but we need to ensure that ANVIL_DEBUG_SAFETY_EXPENSIVE is turned off if we are off
+        private static readonly ScriptDefineDefinition ANVIL_DEBUG_SAFETY_DEFINITION = new ScriptDefineDefinition(ANVIL_DEBUG_SAFETY,
+                                                                                                                  ANVIL_DEBUG_SAFETY_MENU_PATH,
+                                                                                                                  null,
+                                                                                                                  new[]
+                                                                                                                  {
+                                                                                                                      ANVIL_DEBUG_SAFETY_EXPENSIVE
+                                                                                                                  });
+
+        //For ANVIL_DEBUG_SAFETY_EXPENSIVE, we require ANVIL_DEBUG_SAFETY to be on if we're on, but we don't need to turn anything off if we turn off
+        private static readonly ScriptDefineDefinition ANVIL_DEBUG_SAFETY_EXPENSIVE_DEFINITION = new ScriptDefineDefinition(ANVIL_DEBUG_SAFETY_EXPENSIVE,
+                                                                                                                            ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH,
+                                                                                                                            new[]
+                                                                                                                            {
+                                                                                                                                ANVIL_DEBUG_SAFETY
+                                                                                                                            },
+                                                                                                                            null);
+
+
+        [MenuItem(ANVIL_DEBUG_SAFETY_MENU_PATH, true)]
+        private static bool Toggle_Validator_Safety()
+        {
+            return ScriptDefinesToggle.Toggle_Validator(ANVIL_DEBUG_SAFETY_DEFINITION);
+        }
+
+        [MenuItem(ANVIL_DEBUG_SAFETY_MENU_PATH)]
+        private static void Toggle_Safety()
+        {
+            ScriptDefinesToggle.Toggle(ANVIL_DEBUG_SAFETY_DEFINITION);
+        }
+
+        [MenuItem(ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH, true)]
+        private static bool Toggle_Validator_Safety_Expensive()
+        {
+            return ScriptDefinesToggle.Toggle_Validator(ANVIL_DEBUG_SAFETY_EXPENSIVE_DEFINITION);
+        }
+
+        [MenuItem(ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH)]
+        private static void Toggle_Safety_Expensive()
+        {
+            ScriptDefinesToggle.Toggle(ANVIL_DEBUG_SAFETY_EXPENSIVE_DEFINITION);
+        }
+    }
+}
--- a/Scripts/Editor/Debug/AnvilDebugSafetyScriptDefines.cs
+++ b/Scripts/Editor/Debug/AnvilDebugSafetyScriptDefines.cs
@@ -13,13 +13,13 @@ namespace Anvil.Unity.Editor.Debug
         /// Should not be performance intensive.
         /// </summary>
         public const string ANVIL_DEBUG_SAFETY = "ANVIL_DEBUG_SAFETY";
-        
+
         /// <summary>
         /// Script Define for code safety checks that are expensive to determine.
         /// </summary>
         public const string ANVIL_DEBUG_SAFETY_EXPENSIVE = "ANVIL_DEBUG_SAFETY_EXPENSIVE";
-        
-        
+
+
         private const string ANVIL_DEBUG_SAFETY_MENU_PATH = ScriptDefinesToggle.MENU_PATH_BASE + "/" + ANVIL_DEBUG_SAFETY;
         private const string ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH = ScriptDefinesToggle.MENU_PATH_BASE + "/" + ANVIL_DEBUG_SAFETY_EXPENSIVE;
 
@@ -48,28 +48,28 @@ namespace Anvil.Unity.Editor.Debug
         }
 
 
-        [MenuItem(ANVIL_DEBUG_SAFETY_MENU_PATH, true)]
-        private static bool Toggle_Validator_Safety()
-        {
-            return ScriptDefinesToggle.Toggle_Validator(ANVIL_DEBUG_SAFETY_DEFINITION);
-        }
-
         [MenuItem(ANVIL_DEBUG_SAFETY_MENU_PATH)]
         private static void Toggle_Safety()
         {
             ScriptDefinesToggle.Toggle(ANVIL_DEBUG_SAFETY_DEFINITION);
         }
 
-        [MenuItem(ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH, true)]
-        private static bool Toggle_Validator_Safety_Expensive()
+        [MenuItem(ANVIL_DEBUG_SAFETY_MENU_PATH, true)]
+        private static bool Toggle_Safety_Validator()
         {
-            return ScriptDefinesToggle.Toggle_Validator(ANVIL_DEBUG_SAFETY_EXPENSIVE_DEFINITION);
+            return ScriptDefinesToggle.Toggle_Validator(ANVIL_DEBUG_SAFETY_DEFINITION);
         }
 
         [MenuItem(ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH)]
         private static void Toggle_Safety_Expensive()
         {
             ScriptDefinesToggle.Toggle(ANVIL_DEBUG_SAFETY_EXPENSIVE_DEFINITION);
+        }
+
+        [MenuItem(ANVIL_DEBUG_SAFETY_EXPENSIVE_MENU_PATH, true)]
+        private static bool Toggle_Safety_Expensive_Validator()
+        {
+            return ScriptDefinesToggle.Toggle_Validator(ANVIL_DEBUG_SAFETY_EXPENSIVE_DEFINITION);
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Scripts/Editor/Debug/ScriptDefinesToggle.cs
```diff
@@ -0,0 +1,99 @@
...
+            
+            //If the define is already turned on
+            if (definesHashSet.Contains(scriptDefineDefinition.Define))
+            {
+                //Turn it off as well as everything that depends on it
+                definesHashSet.Remove(scriptDefineDefinition.Define);
+                foreach (string dependentDefine in scriptDefineDefinition.DependentDefines)
+                {
+                    definesHashSet.Remove(dependentDefine);
+                }
```
mbaker3, line 67:
```suggestion
                defineHashSet.ExceptWith(scriptDefineDefinition.DependentDefines);
```

Scripts/Editor/Debug/ScriptDefinesToggle.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+            else
+            {
+                //Turn it on as well as everything that it requires
+                definesHashSet.Add(scriptDefineDefinition.Define);
+                foreach (string requiredDefine in scriptDefineDefinition.RequiredDefines)
+                {
+                    definesHashSet.Add(requiredDefine);
+                }
```
mbaker3, line 76:
```suggestion
                definesHashSet.UnionWith(scriptDefineDefinition.RequiredDefines);
```

Scripts/Editor/Debug/ScriptDefinesToggle.cs
```diff
@@ -0,0 +1,99 @@
...
+            Menu.SetChecked(scriptDefineDefinition.MenuPath, isEnabled);
+            return true;
+        }
+        
+        /// <summary>
+        /// Toggle function to turn on or off a given Script Define.
+        /// Call this to toggle a Script Define from the menu.
+        /// </summary>
+        /// <param name="scriptDefineDefinition">The <see cref="ScriptDefineDefinition"/> representing the define</param>
+        public static void Toggle(ScriptDefineDefinition scriptDefineDefinition)
```
mbaker3, line 50: In a simple case I think this works fine but what if two different `ScriptDefineDefinition` instances have the same define as a requirement?

If both are enabled and you disable one the required defines will be removed for the one define that is still enabled.

As we make use of this I see a lot of options clobbering each other's defines and having a hard time tracking it down.

I don't know if there's a way to do this nicely without getting every define to buy into a system. We may be better off just asking developers to enable and dependent conditions with a `#if MY_HIGH_LEVEL_TOGGLE && !MY_DEPENDENCY` condition somewhere in the code that reports the requirement. Enable isn't a problem but cleaning up/disabling dependencies is going to be a manual process I think.

Luckily a git diff will generally catch it easily.

mbaker3 (changes requested):
Mostly minor comments/improvements but take a look at my comment on `ScriptDefinesToggle.Toggle` before addressing anything else. I'm not sure this approach will work.

Scripts/Editor/Debug/ScriptDefinesToggle.cs
jkeon: Ok this was a good point and made me hem and haw a bit over just getting something simple in vs making a fully blown Editor panel that allows for a ton of configuration and rules and error checking.

I built a test case and it ended up being pretty quick to make something more robust but still just a menu item. Full blown panel can come later if we decide to get indulgent. 

I have a define called `BASE` which for example purposes is not owned by us. 

Then a define called `FEATURE_A` which is owned by us and requires `BASE` to be set.
Then a define called `FEATURE_B` which is also owned by us and requires `BASE` to be set. 

Both `FEATURE_A` and `FEATURE_B` appear in the menu because we defined them and said we want them to be there. 

Finally there is another define called `DERIVED` which is not owned by us but both `FEATURE_A` and `FEATURE_B` have said that it is dependent on them. 

In a blank slate, nothing is on.

Go to the menu, toggle `FEATURE_A`. 
The script turns on `FEATURE_A` and also turns on `BASE` because it's a requirement. 
Go to the menu, toggle `FEATURE_B`.
The script turns on `FEATURE_B` and `BASE` is already on. 

Now go into the project settings and manually remove `BASE`. This simulates someone manually messing with things, another script turning it off, or another third party library with their own script define toggles turning off something we rely on. 

The script detects the script define change, realizes that we have `FEATURE_A` and `FEATURE_B` that depended on `BASE` which is now gone and turns them off too. 

On the `DERIVED` side, i think it's a bit of stretch edge case. Why would a third party define require one or more defines in our framework? But still.

Manually go and add `DERIVED` to project settings. Simulating again, someone messing with things, other scripts or frameworks. 

`DERIVED` requires `FEATURE_A` and `FEATURE_B` to be set which both require `BASE` to be set. They aren't, so we get a helpful error in the console telling us to enable those requirements. `DERIVED` is not set.

Finally, assume everyone is set. Then manually remove `BASE`. All of `BASE`s dependent defines are now removed as well. 

The only thing we don't handle is exclusions. Such as if `FEATURE_A` is enabled, `FEATURE_B` must NOT be enabled. Maybe we could add support for that in the future but I feel like the developer of `FEATURE_A` is going to build in some sort of guard in their own code anyway to yell at the developer if both are enabled anyway. Seems like an unlikely scenario given how Script Defines in C# work anyway.

Clobbering should now not be an issue. Everything, (our own defines and existing or third party defines) have requirements and dependencies merged and the toggling logic is explicit when you are making an explicit choice through the menu:

- Enable all requirements when enabling
- Disable all dependants when disabling

But it's lenient when an implicit choice happens through manually editing or a different third party script changes.

- Log to console necessary requirements when enabling
- Disable all dependants when disabling

This leaves you always in a "Will compile" state.

Scripts/Editor/Debug/ScriptDefineDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <see cref="ScriptDefinesToggle"/>
+    /// </summary>
+    public class ScriptDefineDefinition
+    {
+        internal string Define { get; }
+        internal string MenuPath { get; }
+        internal HashSet<string> RequiredDefines { get; }
+        internal HashSet<string> DependentDefines { get; }
+
+        public ScriptDefineDefinition(string define, string menuPath, IEnumerable<string> requiredDefines, IEnumerable<string> dependentDefines)
```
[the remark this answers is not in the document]
jkeon: Done

Scripts/Editor/Debug/ScriptDefineDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+            int dependentCount = DependentDefines.Count;
+            RequiredDefines.UnionWith(other.RequiredDefines);
+            DependentDefines.UnionWith(other.DependentDefines);
+            
+            //If something changed we should validate it's safe
+            if (RequiredDefines.Count == requiredCount
+             && DependentDefines.Count == dependentCount)
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
jkeon: I'm not sure what you wanted to change here but regardless, no longer applicable after my recent changes.

Scripts/Editor/Debug/ScriptDefineDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public ScriptDefineDefinition(string define, string menuPath, IEnumerable<string> requiredDefines, IEnumerable<string> dependentDefines)
+        {
+            Define = define;
+            MenuPath = menuPath;
+            RequiredDefines = requiredDefines != null ? requiredDefines.ToHashSet() : new HashSet<string>();
+            DependentDefines = dependentDefines != null ? dependentDefines.ToHashSet() : new HashSet<string>();
+        }
+
+        internal void MergeIn(ScriptDefineDefinition other)
```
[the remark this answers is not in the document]
jkeon: No longer applicable

Scripts/Editor/Debug/ScriptDefineDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>
+    /// Defines information for use with a Script Define and the behaviour when toggling on and off.
+    /// <see cref="ScriptDefinesToggle"/>
+    /// </summary>
+    public class ScriptDefineDefinition
+    {
+        internal string Define { get; }
+        internal string MenuPath { get; }
+        internal HashSet<string> RequiredDefines { get; }
+        internal HashSet<string> DependentDefines { get; }
```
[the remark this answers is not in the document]
jkeon: We talked about this in another PR in the DOTS repo and auto-getters are the same as readonly fields. 

I have taken to using the pattern of making anything `public`, `protected` or `internal` into auto-getters because then it's easier to use for interfaces in the future.

Scripts/Editor/Debug/ScriptDefineDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+
+            HashSet<string> dupes = new HashSet<string>(RequiredDefines);
+            dupes.IntersectWith(DependentDefines);
+            if (dupes.Count == 0)
+            {
+                return;
+            }
+
+            throw new InvalidOperationException($"Circular reference detected! The script define {Define} requires and depends on {string.Join("," , dupes)}. Please check your {nameof(ScriptDefineDefinition)}s to ensure you have set them up properly.");
```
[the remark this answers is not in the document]
jkeon: Ah neat! I can't use what you suggested because I'm using the `dupes` to output to the error. 

But I did find the whole pattern was wonky to read so just doing:

`IEnumerable<string> dupes = Required.Intersect(Dependent)` is  great!

Scripts/Editor/Debug/ScriptDefinesToggle.cs
jkeon: No longer applicable

Scripts/Editor/Debug/ScriptDefinesToggle.cs
mbaker3: > The only thing we don't handle is exclusions. Such as if `FEATURE_A` is enabled, `FEATURE_B` must NOT be enabled. Maybe we could add support for that in the future but I feel like the developer of `FEATURE_A` is going to build in some sort of guard in their own code anyway to yell at the developer if both are enabled anyway. Seems like an unlikely scenario given how Script Defines in C# work anyway.

That's fine. I agree we don't need this feature.

> - Disable all dependants when disabling

In your example if `FEATURE_A` and `FEATURE_B` are enabled then I disable `FEATURE_B`. `BASE` will remain right? Because `FEATURE_A` still needs it.

Scripts/Editor/Debug/ScriptDefineDefinition.cs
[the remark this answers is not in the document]
mbaker3: My suggestion was to remove the check and early out. Just always do the work.

Scripts/Editor/Debug/ScriptDefinesToggle.cs
jkeon: Correct!

Scripts/Editor/Debug/ScriptDefineDefinition.cs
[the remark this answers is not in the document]
jkeon: Oh... right, that's what the minus signs and red background mean... 🤦

Scripts/Editor/Debug/AnvilDebugSafetyScriptDefines.cs
```diff
@@ -0,0 +1,75 @@
...
+        private static bool Toggle_Validator_Safety()
+        {
+            return ScriptDefinesToggle.Toggle_Validator(ANVIL_DEBUG_SAFETY_DEFINITION);
+        }
+
+        [MenuItem(ANVIL_DEBUG_SAFETY_MENU_PATH)]
+        private static void Toggle_Safety()
+        {
+            ScriptDefinesToggle.Toggle(ANVIL_DEBUG_SAFETY_DEFINITION);
+        }
```
mbaker3, line 61: :bike: I've been putting the validator on the bottom. For consistency these should be but it's not a big deal.

Also, have been putting `_Validator` at the end so the method names more closely match.

Scripts/Editor/Debug/ScriptDefineDefinition.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Anvil.Unity.Editor.Debug
+{
+    /// <summary>
+    /// Defines information for use with a Script Define and the behaviour when toggling on and off.
+    /// <see cref="ScriptDefinesToggle"/>
```
mbaker3, line 9:
```suggestion
    /// See: <see cref="ScriptDefinesToggle"/>
```

Scripts/Editor/Debug/ScriptDefinesToggle.cs
```diff
@@ -0,0 +1,216 @@
...
+                {
+                    continue;
+                }
+                
+                //Ensure that our requirements are present in case some other script or the developer manually removed them
+                foreach (string define in scriptDefineDefinition.RequiredDefines.Where(define => !definesHashSet.Contains(define)))
+                {
+                    LOGGER.Error($"{scriptDefineDefinition.Define} requires that {define} is set, but it's not present! Removing {scriptDefineDefinition.Define} and all dependencies if they were set.");
+                    RemoveDependentDefines(scriptDefineDefinition, definesHashSet);
+                    PlayerSettings.SetScriptingDefineSymbols(NAMED_BUILD_TARGET, definesHashSet.ToArray());
```
mbaker3, line 56: Should this be outside the `foreach` loops?
Otherwise, you're potentially repeatedly setting the define symbols in each iteration of the loop

mbaker3 (dismissed):

mbaker3:
Oops, hit submit too soon. Take or leave comments. Nothing major

Scripts/Editor/Debug/AnvilDebugSafetyScriptDefines.cs
jkeon: Ok sure, i'm not opposed to any of those. Done.

Scripts/Editor/Debug/ScriptDefineDefinition.cs
jkeon: Done

Scripts/Editor/Debug/ScriptDefinesToggle.cs
jkeon: Good catch. Updated

mbaker3 (approved):

PR #114, opened 2023-05-18
Adding an Editor Input Dialog

jkeon:
Unity has a built in Dialog to give you information and buttons but nothing to let the user enter in their own text. 

This doesn't exist today.

We now have a dialog that can take in user input. 

 - None

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Editor/UI/EditorInputDialog.cs
+++ b/Scripts/Editor/UI/EditorInputDialog.cs
@@ -1,4 +1,3 @@
-using System;
 using UnityEditor;
 using UnityEngine;
 
@@ -21,11 +20,9 @@ namespace Anvil.Unity.Editor.UI
         /// <param name="cancelBtnText">What the "Cancel" button should read</param>
         /// <param name="defaultInputText">The default text in the input box if needed</param>
         /// <returns>
-        /// The input text the user entered.
-        /// If the user pressed the "OK" button, this will be whatever they have entered.
-        /// If the user pressed the "Cancel" button, this will be string.Empty.
+        /// The <see cref="EditorInputDialogResult"/>
         /// </returns>
-        public static string Show(
+        public static EditorInputDialogResult Show(
             string title,
             string description,
             string inputTextLabel,
@@ -34,8 +31,6 @@ namespace Anvil.Unity.Editor.UI
             string defaultInputText = null)
         {
             Vector2 maxPos = GUIUtility.GUIToScreenPoint(new Vector2(Screen.width, Screen.height));
-
-            string result = string.Empty;
 
             EditorInputDialog window = CreateInstance<EditorInputDialog>();
             window.Init(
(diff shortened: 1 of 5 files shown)


Scripts/Editor/UI/EditorInputDialog.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        private void CheckKeyboard(Event currentEvent)
+        {
+            if (currentEvent.type != EventType.KeyDown)
+            {
+                return;
+            }
+            
+            // ReSharper disable once SwitchStatementMissingSomeEnumCasesNoDefault
```
mbaker3, line 132: you could probably just add a `default` condition to the switch that called `break` and you wouldn't need this.

Scripts/Editor/UI/EditorInputDialog.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            position = new Rect(mousePos.x, mousePos.y, position.width, position.height);
+
+            Focus();
+        }
+
+        private void Complete(bool isCancel)
+        {
+            OnComplete?.Invoke(isCancel ? string.Empty : m_InputText);
```
mbaker3, line 174: Would it be better to communicate cancel and let the developer read the string off the instance instead?

Scripts/Editor/UI/EditorInputDialog.cs
```diff
@@ -0,0 +1,178 @@
...
+            EditorInputDialog window = CreateInstance<EditorInputDialog>();
+            window.Init(
+                maxPos,
+                title,
+                description,
+                inputTextLabel,
+                okBtnText,
+                cancelBtnText,
+                defaultInputText);
+            window.OnComplete += (resultString) => result = resultString;
```
mbaker3, line 49: Since this method doesn't seem to return until the window closes should we just read the input state off the dialog? No need to have an event.

Scripts/Editor/UI/EditorInputDialog.cs
```diff
@@ -0,0 +1,178 @@
...
+            window.ShowModal();
+
+            return result;
```
mbaker3, line 52: DO we want to communicate cancel vs ok so the shower knows whether to apply the input?

mbaker3 (dismissed):
Looks good. Minor take or leave suggestions

Scripts/Editor/UI/EditorInputDialog.cs
jkeon: I tried and it just stays greyed out and I get a different ReSharper error to disable once instead.

Scripts/Editor/UI/EditorInputDialog.cs
jkeon: Yes, refactored

mbaker3 (approved):

PR #116, opened 2023-09-29
Adding Console Helper for finding first Error in the log

jkeon:
Simple helper utility to get the first error in the Console Window.

Often when developing there will be a ton of logs emitted in the Console Window. Eventually there might be an error. 

To find the first error, you can easily turn off regular logs and warnings so you only have Errors to see it. 

That might be fine but usually reading the regular/warning logs beforehand are helpful to understand the context of the Error. 

There is no way to do this and you may have to scroll manually to try and find the error which can be challenging and annoying.

A simple Tools menu item under `Anvil / Debug / Skip to First Error` exists. 

This will scroll to the first error and select it for you. You can then easily scroll up manually to be able to see the context above. 

- None

 - None

 - [ ] Yes
 - [x] No

### Additional Information

This also exposes a neat feature that we can hijack from how Unity's own Packages `Unity.Rendering`, `Unity.InputSystem` etc are able to access the internals in `UnityEngine.dll` and `UnityEditor.dll`

Unity has exposed special paths for Assembly Definitions called "InternalBridges" as friends of the `UnityEngine.dll` and `UnityEditor.dll` assemblies. 

Since we are unable to create an Assembly Reference for those built in dlls, we can't use the normal method of Assembly Injection. 

Instead we put our code into one of the bridges and we get access that way.

**Engine** - `Unity.InternalAPIEngineBridge.001`
**Editor** - `Unity.InternalAPIEditorBridge.001`

### DIFF
--- a/Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
+++ b/Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
@@ -1,0 +1,88 @@
+using System;
+using System.Reflection;
+using UnityEditor;
+using UnityEngine;
+
+namespace Anvil.Unity.Editor.Debug
+{
+    public static class ConsoleHelper
+    {
+        //Had to figure this out from old Unity decompiled code. Not sure where this is defined today sadly.
+        //But it's unlikely to ever change as so much would have to be changed in the Unity Editor to accomodate.
+        private const int ERROR_FLAG = 512;
+        
+        [MenuItem("Anvil/Debug/Skip to First Error")]
+        internal static void SkipToFirstError()
+        {
+            //If we're not showing errors in the console window, we can't really scroll to it can we?
+            if ((LogEntries.consoleFlags & ERROR_FLAG) == 0)
+            {
+                //Let the user decide...
+                if (EditorUtility.DisplayDialog(
+                    "Incorrect Configuration",
+                    "Errors are not being show in the Console currently. Do you want them to be enabled and try again?",
+                    "Enable",
+                    "Cancel"))
+                {
+                    LogEntries.SetConsoleFlag(ERROR_FLAG, true);
+                }
+                else
+                {
+                    return;
+                }
+            }
+            
+            int errorCount = 0;
+            int warningCount = 0;
+            int defaultCount = 0;
+            LogEntries.GetCountsByType(ref errorCount, ref warningCount, ref defaultCount);
+
+            //No point doing any work if there aren't any errors to scroll to.
+            if (errorCount == 0)
+            {
+                return;
+            }
+
+            Type consoleWindowType = typeof(ConsoleWindow);
+            ConsoleWindow consoleWindowInstance = (ConsoleWindow)consoleWindowType.GetField("ms_ConsoleWindow", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null);
+            ListViewState listViewState = (ListViewState)consoleWindowType.GetField("m_ListView", BindingFlags.Instance | BindingFlags.NonPublic).GetValue(consoleWindowInstance);
+
+            int numLogEntries = LogEntries.GetCount();
+            LogEntry outputLogEntry = new LogEntry();
+
+            LogEntries.StartGettingEntries();
+            for (int i = 0; i < numLogEntries; ++i)
+            {
+                LogEntries.GetEntryInternal(i, outputLogEntry);
+
+                if (!IsError(outputLogEntry.mode))
+                {
+                    continue;
+                }
+
+                Vector2 scrollPos = listViewState.scrollPos;
+                scrollPos.y = i * listViewState.rowHeight;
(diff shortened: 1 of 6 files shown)


Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+
+namespace Anvil.Unity.Editor.Debug
+{
+    public static class ConsoleHelper
+    {
+        // TODO: Future
+        // Should one wish to extend this functionality, adding this as a custom button overlaid on the Console Window
+        // is probably helpful and better than a tools menu.
+        // Additional features would be the ability to keep track of which error we are on and be able to go next/back
+        // as well as first and last.
```
mbaker3, line 14: make an issue number

Scripts/Editor/AssemblyInjections/UnityEditor/Unity.InternalAPIEditorBridge.001.asmdef
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Unity.InternalAPIEngineBridge.001",
```
mbaker3, line 2: hah, wow ok. More magic Unity names...
Cool ability though!

Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        // is probably helpful and better than a tools menu.
+        // Additional features would be the ability to keep track of which error we are on and be able to go next/back
+        // as well as first and last. 
+        
+        //Had to figure this out from old Unity decompiled code. Not sure where this is defined today sadly.
+        //But it's unlikely to ever change as so much would have to be changed in the Unity Editor to accomodate.
+        private const int ERROR_FLAG = 512;
+        
+        [MenuItem("Anvil/Debug/Skip to First Error")]
+        internal static void SkipToFirstError()
```
mbaker3, line 21: A skip to next error would be nice as well but we can do later if it's not a super quick addition.
Often when compiling+running you get some native dispose errors from the previous run instance's crash. In those cases the first error(s) are unrelated to a runtime issue.

Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
```diff
@@ -0,0 +1,94 @@
+using System;
+using System.Reflection;
+using UnityEditor;
+using UnityEngine;
+
+namespace Anvil.Unity.Editor.Debug
```
mbaker3, line 6: Just noticed this and we should push this off to another PR but should we move all of this namespace to `Debugging` to avoid conflict with the `Debug` class?

That's what we did in the runtime code.

mbaker3 (changes requested):
Looks good. Just one cleanup item
A couple nice to have asks as well.

Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
jkeon: Yep, done!

Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
jkeon: Done

Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
jkeon: Yeah it'll have to be later, it's not super difficult I don't think, but will require a bit of effort.

Scripts/Editor/AssemblyInjections/UnityEditor/Unity.InternalAPIEditorBridge.001.asmdef
jkeon: Yeah odd way to do it, but after many a google search and travelling through the forums.... I finally found it

Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
mbaker3: Sorry I meant all of the classes under this namespace `ScriptDefinesToggle`, `ScriptDefineDefinition`, etc...

Let's do that as part of a new PR though since this one is good to go.

mbaker3 (approved):

Scripts/Editor/AssemblyInjections/UnityEditor/ConsoleHelper.cs
jkeon: Cool, tracked in issue here: https://github.com/decline-cookies/anvil-unity-core/issues/118

PR #123, opened 2024-03-15
Hotfix double destroy

jkeon:
If a `GameObject` is destroyed via GameObject.Destroy, it will then trigger the Dispose which will then trigger the Destroy again. 

This was discovered by having an `AbstractAnvilMonoBehaviour` executing in the Editor via the `ExecuteAlways` attribute and seeing how Unity would call Destroy on it resulting in the double destroy. 

- None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Core/AbstractAnvilMonoBehaviour.cs
+++ b/Scripts/Runtime/Core/AbstractAnvilMonoBehaviour.cs
@@ -24,8 +24,6 @@ namespace Anvil.Unity.Core
         /// </summary>
         public bool IsDisposing { get; private set; }
 
-        private bool m_IsDestroying;
-
         private Logger? m_Logger;
 
         /// <summary>
@@ -53,6 +51,11 @@ namespace Anvil.Unity.Core
         /// </summary>
         public void Dispose()
         {
+            Dispose(true);
+        }
+
+        private void Dispose(bool shouldDestroyGameObject)
+        {
             if (IsDisposing || IsDisposed)
             {
                 return;
@@ -61,14 +64,13 @@ namespace Anvil.Unity.Core
             IsDisposing = true;
             DisposeSelf();
 
-            if (!m_IsDestroying)
+            if (shouldDestroyGameObject)
             {
 #if !UNITY_EDITOR
-            Destroy(gameObject);
+                Destroy(gameObject);
 #else
                 //If we're running in the Editor via the ExecuteAlways attribute, we're not allowed to call Destroy
-                if (Application.isEditor
-                 && !Application.isPlaying)
+                if (!Application.isPlaying)
                 {
                     DestroyImmediate(gameObject);
                 }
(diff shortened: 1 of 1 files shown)


Scripts/Runtime/Core/AbstractAnvilMonoBehaviour.cs
```diff
@@ -58,17 +60,39 @@ public void Dispose()
...
             IsDisposing = true;
             DisposeSelf();
+
+            if (!m_IsDestroying)
+            {
+#if !UNITY_EDITOR
             Destroy(gameObject);
+#else
+                //If we're running in the Editor via the ExecuteAlways attribute, we're not allowed to call Destroy
+                if (Application.isEditor
```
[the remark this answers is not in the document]
jkeon: Good point, removed.

Scripts/Runtime/Core/AbstractAnvilMonoBehaviour.cs
```diff
@@ -58,17 +60,39 @@ public void Dispose()
...
-            if (IsDisposing || IsDisposed)
+            if (IsDisposing
+             || IsDisposed
+             || m_IsDestroying)
             {
                 return;
             }
+            m_IsDestroying = true;
 
             Dispose();
```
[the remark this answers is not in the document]
jkeon: Invalid because your overall suggestion was a better idea.

jkeon:
> I've left a couple comments on the current implementation but I'm wondering if we'd be better served by changing up the pattern a bit to avoid needing the extra field.
> 
> If we change the signature of Dispose to Dispose(bool skipDestroy = false) then we could bypass the destroy call when coming from the OnDestroy() stack without the need to carry a field on the type.

Yeah this was a better idea. Changed to reflect that. We still have the original signature so we match the `IDisposable` interface but then we just pipe to the param version.

mbaker3 (approved):

Scripts/Runtime/Core/AbstractAnvilMonoBehaviour.cs
```diff
@@ -50,6 +50,11 @@ protected virtual void Awake()
         /// Will Destroy the attached <see cref="GameObject"/> if it has not been destroyed yet.
         /// </summary>
         public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        private void Dispose(bool shouldDestroyGameObject)
```
mbaker3, line 57: Might make sense to make this public some day if you only want to dispose the specific behaviour but we should wait until the need actually comes up.

We'd need a `RemoveComponent` path anyway.

jkeon: Yeah that's fair. Interesting that in all these years that's never been a pattern we've done.

mbaker3: Yea, the whole dispose pattern is a bit of an opinionated choice.

On one hand you could say that we should have `Dispose` at all and developers should just use the engine commands `Object.Destroy` and `MonoBehaviour.RemoveComponent`.

On the other hand, it's nice to implement the `IDisposable` pattern and, personally, I prefer to call `Dispose` vs using Unity's various ways of destroying objects.

jkeon: Agreed!

PR #124, opened 2024-06-17
Added `ScaleAroundPoint` and `RotateAroundPoint` transform extensions

tjvezina:
Added `ScaleAroundPoint` and `RotateAroundPoint` transform extensions

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Editor/Tests/Core/Util/TransformExtensionTests.cs
+++ b/Scripts/Editor/Tests/Core/Util/TransformExtensionTests.cs
@@ -1,9 +1,11 @@
 using Anvil.Unity.Core;
 using NUnit.Framework;
 using UnityEngine;
+using UnityEngine.TestTools.Utils;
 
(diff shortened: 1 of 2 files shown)


mbaker3 (changes requested):
No rush, but can we get tests for these additions?
I only ask because `TransformExtension` currently hast test coverage.

Otherwise, looks good!

tjvezina:
Tests added—the `RotateAroundPoint` test was suffering a rounding error, giving false negative results, but a custom Vector3 comparer with `10e-7` margin of error fixed it.

mbaker3 (approved):
Looks good. Thanks!

> Tests added—the RotateAroundPoint test was suffering a rounding error, giving false negative results, but a custom Vector3 comparer with 10e-7 margin of error fixed it.

I've run into this before and we have a precision constant defined somewhere but I like this solution better.

PR #127, opened 2024-07-06
Add PhysicsUtil

mbaker3:
Add a method to extrapolate the state of a rigid body by a delta without needing to step Unity's physics system.

None

The method approximates the clean movement of a rigid body accounting for gravity, linear drag and angular drag. Does not account for collisions or any changes applied through the FixedUpdate of MonoBehaviours. However the latter can be accomplished by providing an onStep callback and calling the same logic that any FixedUpdate behaviours do.

Note: That values aren't re-read off of the RigidBody so all changes in onStep must make changes to the state parameters passed into the ExtrapolateState method.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Physics.meta
+++ b/Scripts/Runtime/Physics.meta
@@ -1,0 +1,3 @@
+fileFormatVersion: 2
+guid: 72fad2898ac24c5a9b80a816db388516
+timeCreated: 1720231943
--- a/Scripts/Runtime/Physics/PhysicsUtil.cs
+++ b/Scripts/Runtime/Physics/PhysicsUtil.cs
@@ -1,0 +1,116 @@
+using System;
+using Unity.Mathematics;
+using UnityEngine;
+
+namespace Anvil.Unity.Physics
+{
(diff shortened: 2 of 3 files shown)


jkeon (approved):
Looks good. 

I hemmed and hawed over being more explicit with our warning on these functions that we're simplifying what's happening. This won't take into account any other forces acting on the object or it's mass for example, but I think the current note of:

```
/// Note: Does not account for collisions or trigger <see cref="FixedUpdate"/> on the body.
///       For a more accurate extrapolation <see cref="UnityEngine.Physics.Simulate"/> is more appropriate.
```
is enough.

mbaker3:
> I hemmed and hawed over being more explicit with our warning on these functions that we're simplifying what's happening. This won't take into account any other forces acting on the object or it's mass for example

Yea, I don't think mass applies to drag so I don't believe there's any simplification we're making in our math. It would be up to the developer to account for mass, where appropriate, in their own `onStep` execution.

PR #133, opened 2025-03-22
BoundUtils - Add Initial Utils

mbaker3:
Add
 - CreateBoundingVolume
 - EdgeLookup
 - GetCorners
 - IntersectPlane
 - IntersectPlane(out intersections)

Some of these are extension methods. However, It seems a bit redundant to have `BoundsUtil` and `BoundsExtension`.
I can split them up if people want or have a better idea of how to handle this moving forward.

 - None

 - New utility methods to help working with bounds

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Core/Util/BoundsUtil.cs
+++ b/Scripts/Runtime/Core/Util/BoundsUtil.cs
@@ -1,0 +1,175 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Unity.Mathematics;
+using UnityEngine;
+
+namespace Anvil.Unity.Core
+{
+    /// <summary>
+    /// A collection of utility methods to supplement the <see cref="Bounds"/> object.
+    /// </summary>
+    public static class BoundsUtil
+    {
+        /// <summary>
+        /// A collection of all edges as indices of the results of <see cref="GetCorners"/>.
+        /// </summary>
+        public static readonly int2[] EdgeLookup = new[]
+        {
+            new int2(0, 1),
+            new int2(0, 2),
+            new int2(0, 4),
+            new int2(1, 3),
+            new int2(1, 5),
+            new int2(2, 3),
+            new int2(2, 6),
+            new int2(3, 7),
+            new int2(4, 5),
+            new int2(4, 6),
+            new int2(5, 7),
+            new int2(6, 7)
+        };
+
+        /// <summary>
+        /// Creates <see cref="Bounds"/> from a collection of arbitrary points
+        /// </summary>
+        /// <param name="points">The points to consider.</param>
+        /// <returns><see cref="Bounds"/> containing the points provided.</returns>
+        public static Bounds CreateBoundingVolume(params Vector3[] points)
+        {
+            float3 min = float.MaxValue;
+            float3 max = float.MinValue;
+
+            float4 x = new float4();
+            float4 y = new float4();
+            float4 z = new float4();
+
+            //TODO: #99 - Profile and evaluate whether this can be improved.
+            // - Do math.min(i + 2, points.Length) instead of (i + 1) % points.Length to avoid potential cache invalidation
+            // - Strides of 4 but do a separate math.min(batchMinX, minX) call
+            // - Is it even worth batching?
+            for (int i = 0; i < points.Length; i += 3)
+            {
+                Vector3 point1 = points[i];
+                // wrap around if the batch size is not a multiple of points.Length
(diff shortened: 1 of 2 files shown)


Scripts/Runtime/Core/Util/BoundsUtil.cs
```diff
@@ -0,0 +1,175 @@
...
+            new int2(5, 7),
+            new int2(6, 7)
+        };
+
+        /// <summary>
+        /// Creates <see cref="Bounds"/> from a collection of arbitrary points
+        /// </summary>
+        /// <param name="points">The points to consider.</param>
+        /// <returns><see cref="Bounds"/> containing the points provided.</returns>
+        public static Bounds CreateBoundingVolume(params Vector3[] points)
```
jkeon, line 37: This function is mixing `math` and `Mathf` data structures. 

We seem to be doing as much as we can with `math` and then ultimately converting back to Unity's `Bounds` so I would suggest this takes in `float3` instead of `Vector3`

And use `float3` down on lines 52 to 55

You can still pass `Vector3`'s in and it'll implicitly convert them.

jkeon (changes requested):
Minor change but otherwise looks good. 

There are a bunch of TODOs which are totally fair, do we want to tie those to Issues? 

With regards to the `Util` vs `Extension` I thought we already were splitting them up but I took a look and can't find an example where we are. 

I think it makes sense to split them from a purist standpoint but I agree that it seems redundant. What I do see happening though is we'll start an make an `XExtension` and a `YUtil` that have only extensions methods or static methods respectively and then we'll add another method that doesn't fit the mould and the class will live on with it's first given name. Which makes me think that maybe we need a different term that encompasses both?

Scripts/Runtime/Core/Util/BoundsUtil.cs
mbaker3: This is a near copy of the same `RectUtil.CreateBoundingRect` method. I agree that the mixing is a bit weird but there is rationale.

As it's written right now there's no casting penalty. Most of the time you'll be getting a collection of `Vector3` points from Unity and feeding it into this method (Ex mesh vertices). That's the format you're coming into this method with.

Within the method we never pay a casting penalty as we read out everything component-wise.

Finally, bounds is based in the old Unity math and the struct we're stuck with. Here is the one spot we're paying a penalty with two `float3` to `vector3` casts.

If we took in `float3[]` params then we'd be casting many more times.

> You can still pass Vector3's in and it'll implicitly convert them.

This doesn't work when they're wrapped in collections. You'd have to run `.Select<Vector3>()` over them.

mbaker3:
> There are a bunch of TODOs which are totally fair, do we want to tie those to Issues?

I was hoping you'd miss that! I don't have time now but I'll do that before merging.

> 

> With regards to the `Util` vs `Extension` I thought we already were splitting them up but I took a look and can't find an example where we are.
> 
> I think it makes sense to split them from a purist standpoint but I agree that it seems redundant. What I do see happening though is we'll start an make an `XExtension` and a `YUtil` that have only extensions methods or static methods respectively and then we'll add another method that doesn't fit the mould and the class will live on with it's first given name. Which makes me think that maybe we need a different term that encompasses both?

Yea, that's where I was stuck too. Moving away from Util and Extension defies convention general C# convention. In .NET they have the split but from what I've seen they have enough methods in both to make it not seem weird. I'm sure there's a deeper rational for how/why they choose to make any given method an extension vs Util too.

jkeon (dismissed):
> I was hoping you'd miss that! I don't have time now but I'll do that before merging.

Lol, this is Anvil, fine tooth comb time :)

> Yea, that's where I was stuck too. Moving away from Util and Extension defies convention general C# convention. In .NET they have the split but from what I've seen they have enough methods in both to make it not seem weird. I'm sure there's a deeper rational for how/why they choose to make any given method an extension vs Util too.

I think if you have Extension methods in the `Util` class, you don't care because you never reference the class itself in your code. 

Ex. `myBounds.GetCorners()`

But if it was a static method on an `Extension` class, then it feels weird because you'd have to write it out.

Ex. `BoundsExtention.GetCorners(myBounds)` vs `BoundsUtil.GetCorners(myBounds)` 

So unofficially, I think if we just have Extension methods, use `Extension`, if we have both use `Util`.

mbaker3:
@jkeon issues added to TODOs. Ready for approval.

jkeon (approved):

PR #137, opened 2025-06-12
Add GizmoUtils

mbaker3:
A set of gizmo utilities that expand on Unity's offerings.

Add the ability to draw a wire quad gizmo with optional normal

 - None

 - None

 - [ ] Yes 
 - [x] No


mbaker3:
Going to rename `GizmoUtils` to `GizmoUtil` but need to get working project into a better state first.

PR #148, opened 2026-01-16
QuitState - Fix double init

mbaker3:
In editor the quit state could be initialized multiple times causing double OnQuitting events to fire.

`QuitState.OnQuitting` will fire twice on quit. If domain reload is disabled the event will fire an additional time for each consecutive time the project is run without re-compiling.

`QuitState.OnQuitting` will fire a single time per play session.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Good catch!

